PR #5, opened 2023-10-17
Add additional connection_update plugin. Add testing.

jamshale:
Add an additional plugin `connection_update` and unit testing coverage for both currently included plugins.

There are integration tests but both the current plugins don't require much integration testing.

As discussed the automation script will be included in another task. The current goal is to get all the valid plugins from traction into this repo and well tested for the base moving forward.


CONTRIBUTING.md
jamshale, line 1: Not exactly sure what to put in this file yet. Thinking about it.

basicmessage_storage/pyproject.toml
```diff
@@ -55,18 +55,17 @@ testpaths = "basicmessage_storage"
 addopts = """
     --quiet --junitxml=./test-reports/junit.xml
     --cov-config .coveragerc --cov=basicmessage_storage --cov-report term --cov-report xml
-    --ruff
```
jamshale, line 58: This was throwing errors for me. I actually need to figure out how to do the linting still. Right now i'm just using the ruff vs code extension and not sure if all the rules are being met.

basicmessage_storage/setup.cfg
```diff
@@ -1,10 +1,2 @@
-[flake8]
```
jamshale, line 1: Didn't think this was needed anymore as we are using `ruff` ?

configs/multitenant-plugins.yml
jamshale, line 1: Was going to add a basic single tenant config example as well. I also don't know how far we want to take example config files. Seems somewhat out of scope for this repo.

CONTRIBUTING.md
esune: Could take inspiration from ACA-Py's [CONTRIBUTING.md](https://github.com/hyperledger/aries-cloudagent-python/blob/main/CONTRIBUTING.md)

configs/multitenant-plugins.yml
esune: I think I agree, plugin documentation should be covered in ACA-Py already (I think).

updater.sh
```diff
@@ -0,0 +1,119 @@
+#!/usr/bin/bash
```
esune, line 1: If you haven't already: following [this](https://github.com/hyperledger/aries-cloudagent-python/blob/main/scripts/run_docker) example might help having a cross-OS compatibile script.

esune (commented):
This looks quite good, thank you @jamshale! I might try to run things on my machine, in the meantime tagging @dbluhm, @shaangill025 and @usingtechnology for input/review since they have more plug-in development experience than I do.

A couple of thoughts:
- unsure about having `updater.sh` require a specific OS architecture. We might want to dig deeper into making it cross-platform, but it is a good start either way.
- I am wondering if, for testing, having plug-ins added/built into the image in a more randomized way to prevent internal dependencies - unless required - and situations such as "it has to be in this sequence or it won't work" could be something we should think of?

updater.sh
dbluhm, line 1: How is this file intended to be used? I'm not too keen on maintaining dense bash scripts in this repo, if we can avoid it, to be honest :sweat_smile:

Dockerfile
dbluhm, line 1: In the future, I think I'd like to see dockerfiles moved to inside of specific plugins only. I don't know that it makes sense for there to be a top level dockerfile since this repo will house plugins that someone wouldn't necessarily need/want to deploy together.

As an alternative, we could instead include a set of example dockerfiles that could aggregate plugins from this project into a single image?

I'm not against merging these changes right now though.

usingtechnology: Yeah, we could lose this file. or simply add something to documentation. When I added this, it was intended more as documentation for adding a plugin to an external image. Perhaps too confusing and plugin usage and deployment best left as pure documentation.

updater.sh
jamshale: There will only be one. 

It allows populating new plugins with everything they need for integration testing and devcontainer development. Along with updating things for all plugins at once such as the launch.json file and whatnot.

The only reason I want to have this is I feel like this repo will be very challenging to maintain when there is many plugins and adding new plugins with all the infrastructure is a giant pain.

This is what this script is trying to solve.

Dockerfile
jamshale: I do like it as an example of an agent with multiple plugins. But yes it doesn't need to exist.

updater.sh
jamshale: That is a good idea. I should have thought of that. It won't take long to copy the logic.

In my opinion this is critical. I think it becomes pretty obvious when you go to add a new plugin, and think about there being dozens of plugins being contributed by different people. Even with comprehensive readme the instructions are going to be complicated and being able to manage plugins globally is going to become time consuming and error prone.

This doesn't need to be a part of this PR. I could remove it and bring it back as a python script in another PR so it's more clear what was added as a part of plugins and testing and keep it separate from the script.

Dockerfile
usingtechnology: Let's drop the top-level Dockerfile and add an issue for documentation. We do need to let consumers know how to leverage the plugins and include them in their deployments.

updater.sh
jamshale: @usingtechnology Sounds good. I'll do that and update the PR without it.

jamshale:
Open again. This PR will just add `connection_update` plugin and testing and address some comments.
The script will return as another task later on. Still working on testing the next plugin to be added `multitenant_provider`.

usingtechnology (changes requested):

basicmessage_storage/plugin-config.yml.sample
```diff
@@ -1 +1,31 @@
-# document how this plugin is configured and provide decent default settings
\ No newline at end of file
+auto-provision: true
```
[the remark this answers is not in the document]
jamshale: Gotcha. Ya i'll remove them.

usingtechnology (approved):

PR #17, opened 2023-10-23
add multienant_provider plugin

jamshale:
Add unit tests and integration tests


usingtechnology (changes requested):
Like what you are doing here... yeah, lots of super-tricky things to figure out for testing; great work getting it figured out.

One change I'd like to see is in the integration tests. Can we add something that illustrates the tenant getting multiple valid tokens? I think it would be a good thing to test and a good use case to show how a tenant would get their tokens and that they can all be valid.

jamshale:
Thanks for the review. Sounds good. I'll work on expanding the integration tests for that scenario.

jamshale:
Added a couple integration tests for getting multiple tokens from one wallet. Fixed the linting to match the requirements. I did add some documentation rules to the ignore list that I felt were a bit overkill. I can add them back if they are actually wanted.

basicmessage_storage/pyproject.toml
```diff
@@ -43,7 +43,7 @@ ignore = [
     "D203", "D204", "D213", "D215", "D400", "D401", "D404", "D406", "D407",
     "D408", "D409", "D413",
     "D202", # Allow blank line after docstring
-    "D104", # Don't require docstring in public package
+    "D101", "D102", "D104", "D106", "D107",
```
usingtechnology, line 46: applies to all... should we deviate from ACA-Py itself? i think these rules should be the same. even if you put in placeholder comments with TODO and they could get assigned to the creators to fill in properly.

jamshale: Oh. I didn't realize these were the same as aca-py. I'll change it back and update.

jamshale:
Reverted the ruff configuration doc rules and updated the plugin with the required docs.

usingtechnology (approved):

usingtechnology (approved):
This is great - awesome work @jamshale !

PR #18, opened 2023-10-24
Add workflow to ensure plugins follow liniting rules

jamshale:
Adds a push to main workflow which navigates to the root of each plugin. Installs dependencies and then runs the lint check.

Updates the existing plugins to conform to the current rules.

Installing the dependencies for each plugin. It could just install dev dependencies but i'm planning to add unit tests to the same workflow, which will need all dependenices. The unit testing will be built off of this workflow so installing the extra dependencies shouldn't matter.

### DIFF
--- a/.github/workflows/push.yaml
+++ b/.github/workflows/push.yaml
@@ -6,7 +6,7 @@ on:
 
 jobs:
   lint:
-    name: "Perform lint checking with ruff"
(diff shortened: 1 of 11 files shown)


jamshale:
Re-opening from a different branch with additional work. This PR won't let me change the branch I want to merge from.

PR #19, opened 2023-10-26
Feat/15 add workflow that performs lint checks

jamshale:
Creates a github workflow that install poetry, navigates to each plugin, installs and caches dependencies. Then the linting job navigates into each plugin a performs the lint checks declared in the plugins root toml file.

Fixes a few linting issues with the existing plugins.

This workflow currently runs every branch commit. Might be overkill. Could make it for only pull requests or possibly a pre commit hook if desired.

### DIFF
--- a/.github/workflows/push.yaml
+++ b/.github/workflows/push.yaml
@@ -1,0 +1,54 @@
+name: Prepare
+on:
+  push:
+    branches:
+      - "**"
+
+jobs:
+  lint:
+    name: "Code quality checks"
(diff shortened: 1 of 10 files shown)


usingtechnology (approved):
LGTM, like the documentation in the workflow.
Could be an issue importing plugins that do not use ruff but we can make it a standard. Certainly makes it easier for other devs to jump in if we start laying down some standards.  We can keep an eye on tweaking when this is run and adjust accordingly.

PR #20, opened 2023-10-30
Add unit tests to push workflow

jamshale:
This will run unit tests and lint checking in their own workflow and only for plugins that have changes. It has been changed to only run on PR's and push to main.


dbluhm:
Do we really want to run unit tests whenever there's a push to any branch? I can see main and running them on PRs of course but running them on every push feels unnecessary.

jamshale:
I think I'll wait until I get the integration test PR finished and then move these tests to that action to only run on changes as well.

jamshale:
This has been updated.

There was problems with the redis_events testing. I needed to change the integration tests to become an internal network and the pytests weren't locating the plugin root. The easiest way to fix this was to change the sys.path for the pytest init file.

PR #25, opened 2023-10-30
Feat/21 run integration tests on pr

jamshale:
Will run integration tests for plugins with changes on PR creation or edit.

### DIFF
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -1,0 +1,54 @@
+name: Pull Request - Integration Tests
+on:
+  pull_request:
+    types: [opened, edited, synchronize, reopened, ready_for_review]
+    branches:
+      - "**"
+
+jobs:
+  integration-tests:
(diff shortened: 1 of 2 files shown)


dbluhm:
Since this repo will contain many plugins, it doesn't seem ideal for literally all the plugin's integration tests to run for an update to just a single plugin. Additionally, given that integration testing generally spins up containers and services, integration testing can be quite slow; running each set serially will become tedious to wait for.

jamshale:
This has a script command in the github action to determine which plugins have changes and only run those integration tests now.

dbluhm:
@jamshale can we touch a file in one of the plugins just to see the action running?

jamshale:
@dbluhm I added a commit for a small change to multitenant provider plugin and pushed it. You can see the action running only those integration tests now.

dbluhm:
Nice! Looks good!

PR #29, opened 2023-11-07
Add redis events plugin

jamshale:
Adds the redis event persistent queue plugin to the repo.

I wanted consistency with other plugins and for all the code to be contained within the plugin definition so i restructured the directory slightly. I didn't make any code changes other than updating the imports.

The demo was broken. I'm still trying to bring it back but it's been more challenging than I thought and I might make another ticket for this. The load tests in the docs directory were also completely broken so I removed them as well. It would be nice to bring them back.

I updated the integration tests to include the relay service and added a few tests. I still think this should be expanded further, but I think it's good enough for now. In some ways it's better than having the demo imho. I added the basicmessage plugin to the integration test agent to help make sure the messages were received. (I think this is ok because they are integration tests)

I left Shaanjot's name as the author on everything but I removed the inner service pyproject files and combined them together so I changed them slightly.

Updated the pydocs to match the ruff requirements and fixed a few unit tests.


redis_events/setup.cfg
dbluhm, line 1: Minor comment, we should move this to pyproject.toml

redis_events/setup.py
dbluhm, line 1: Minor comment, I believe this file can be removed now

dbluhm (approved):
This looks good to me. Left a couple of super minor comments.

For the included diagrams in png format, it would be kinda nice to see those be turned into https://mermaid.js.org diagrams. That way they can be inline to the relevant markdown document and easily updated in the repo (no need to recreate it in another program). I mostly say this just to seed future work; again, I'm happy with this PR as is.

Example of using mermaid in github:

```mermaid
sequenceDiagram
    Alice->>Bob: Hello Bob, how are you ?
    Bob->>Alice: Fine, thank you. And you?
```

jamshale:
@dbluhm Sounds good. I can make those updates. I would like to avoid extra image files as well. Except for that one complicated design image.

PR #38, opened 2023-11-15
Kafka events

dbluhm:
This PR adds the kafka-events plugin from https://github.com/sicpa-dlab/aries-acapy-plugin-kafka-events.

I open this PR on behalf of the SICPA team who have generously agreed to contribute this plugin both to HL and to the plugins repo.

Thank you SICPA team!

This PR includes the history of the plugin. I find this valuable but I'm open to thoughts on this approach.


swcurran:
@jamshale can you please do the “mechanical” review of the PR?  Are all the pieces needed by the plugins guidelines to be added to the repo.

@dbluhm — I see that the DCO is not complete.  Thoughts on what to do about that?

dbluhm:
Indeed, I'll work with the SICPA to get those commits corrected. On the mechanical points, there are some small tweaks to make (renaming a folder, adding ruff, etc.)

jamshale:
This looks good. It seems very similar to the conversion I did from `aries-acapy-plugin-redis-events`. I would like it to change structure to follow what redis_events is like if possible. The integration tests folder will need to be renamed for the tests to run correctly and I think the workflow files could be removed. There's definitely a bit of work to make it consistent with the other ones.

I've just done almost the same thing so I could do this a well if we wanted, but I wouldn't get it done before I go on vacation.

It is really handy to have the commit history. It would have helped show what I did with the redis_events plugin. I wasn't sure how I could do that.

jamshale:
I think I'll try and update this to try and get the unit, linting, and integration tests working. Then I'll create another ticket to adjust the structure as another task.

dbluhm:
@jamshale quick warning that the DCO issues will require a history rewrite to correct. I think you should be able to just rebase any changes you make but wise to keep that in mind :sweat_smile: As I already mentioned, I'll work with the SICPA team on getting the commits missing a sign-off taken care of but your contributions on those things (unit and int testing, etc) would be greatly appreciated!

jamshale:
I changed the structure of the plugin and adjusted a few files. I added docs for ruff linting and changed the linting/testing configs. The integration tests and unit tests and linting checks are now working. I removed some files related to github actions and pre commit linting checks. Some of these would be nice to have but at a repo level.

I think this could be merged now when the DCO check is fixed. There are several things I would still like to change/update before this plugin is considered finished.

- Dependencies in general and acapy dependency specifically should be updated to the level of the other plugins. 
- It'd be nice to have the Dockerfiles updated to be consistent with the existing plugins.
- Some documentation needs to be updated still after changing the structure.
- Would like more unit testing in general. Hopefully the integration tests catch things, but I've already seen unit tests be beneficial in catching unknown breaking changes in acapy.

I'll see if I can get anymore done today. Otherwise we can merge it as is and I can finish these things when I'm back from vacation.

dbluhm:
@ryjones seeking advice on the best way to handle the DCO issues on this PR. I am in contact with the contributors of the affected commits and they are willing to consider those commits "signed off" according to the DCO. The authors are not, however, in an ideal position to rewrite the history of this branch to add those sign-off messages. Is this a case where using third party remediation would be appropriate? https://github.com/dcoapp/app#third-party-remediation-support

dbluhm:
@ryjones, unfortunately, Luis Gomez is not with SICPA anymore. Can @chumbert sign off the commits on his behalf?

ryjones:
@dbluhm yes - in fact, one representative from SICPA can sign them all off.

dbluhm:
@jamshale @swcurran This is ready for final review!

PR #41, opened 2023-11-17
Repo Management Script

jamshale:
This contains a script for creating scaffolding for a new plugin, managing global dependencies and configuration and common development files.

It's quite a bit simpler than my first attempt. It still contains a new folder where the global files are stored `plugin_globals`, but it doesn't require any additional files from the plugins themselves. It will combine the dependencies and conflicts will be overridden by the global ones. Config sections will just be overriden by the globals (no merging).

There isn't support for plugins to have extra sections currently. They will just be removed. I'm not sure if we want to let plugins be able to do this anyway. Will make things harder to maintain. 

Files in the existing plugins got adjusted slightly. I think these changes could be safely ignored.


jamshale:
I moved main to a function and also added more documentation and tried to improve maintainability with some code changes.

PR #47, opened 2023-12-01
OpenID4VCI Plugin

dbluhm:
This PR adds a plugin implementing the OpenID4VCI protocol (Issuance only -- future work will address OpenID4VP and I'm not sure if that should be in this plugin or a separate one). See the README for details on the implementation.

A couple of things to call out:

- This Plugin Implements Draft 11 of OpenID4VCI (https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0-11.html).
- We have been testing against the Sphereon Wallet. Kudos to the Sphereon team for their work on that!
- To interoperate with Sphereon, we needed to add support for `did:jwk` and verifying `ES256K` signatures. This implementation is within this plugin right now (as a resolver and a helper method). We plan to contribute this to ACA-Py Core.
- We plan to have automated tests against AFJ's OpenID4VCI package as soon as PR https://github.com/openwallet-foundation/agent-framework-javascript/pull/1639 is merged and released.
- Not implemented (yet):
  - `ldp_vc`, `sd_jwt_vc`
  - Authorization Code Flow
  - Only signature suite supported by ACA-Py for jwt-vc right now is `EdDSA`
  - GET /.well-known/openid-configuration
  - GET /.well-known/oauth-authorization-server
  - Batch Credential Issuance
  - We're limited to DID Methods that ACA-Py supports for issuance (more can be added by Plugin, e.g. DID Web); `did:sov`, `did:key`
  - Multitenancy Support; we need to figure out the right way to identify a sub-wallet from OpenID4VCI endpoints.


dbluhm:
@swcurran We'd like to demo this at the next ACA-PUG call!
@jamshale Quick ACK that we have some structural things to adjust to meet the expectations of the GHA (and some failing unit tests, right this minute)

swcurran:
Definitely want to see the demo — great stuff!  Looking forward to it!

jamshale:
@dbluhm I think the unit tests may be failing just because the integration test folder needs to be renamed from int --> integration or it tries to run the integration tests as unit tests.

oid4vci/pyproject.toml
```diff
@@ -0,0 +1,88 @@
...
+    "D417", "C901",
+]
+
+line-length = 90
+
+[tool.ruff.per-file-ignores]
+"**/{tests}/*" = ["F841", "D", "E501"]
+
+[tool.pytest.ini_options]
+testpaths = "basicmessage_storage"
```
jamshale, line 55: Copy and paste error here. I think changing this and the integration tests folder name would get the unit tests succeeding.

dbluhm:
Currently failing unit tests are legit failures; we'll get them taken care of (probably Monday :smile:)

dbluhm:
We'll have more developments on this plugin coming soon but this PR is ready for review!

WadeBarnes (commented):
@dbluhm, Have a look at the way we're updating the ACA-Py dependencies over [here](https://github.com/hyperledger/aries-acapy-plugins/pull/49).  I'd recommend updating before the PR is merged.

jamshale (approved):

dbluhm:
> @dbluhm, Have a look at the way we're updating the ACA-Py dependencies over [here](https://github.com/hyperledger/aries-acapy-plugins/pull/49). I'd recommend updating before the PR is merged.

Good point! Updates pushed implementing the ACA-Py as an optional dependency change adopted in other plugins.

oid4vci/demo/frontend/Dockerfile
```diff
@@ -0,0 +1,23 @@
+# Use an official Node.js runtime as the base image
+FROM node:14
```
esune, line 2: Node 14 has been past its EOL for a while now, any reason not to update it to a newer version?

esune (commented):
Looks good, thanks @dbluhm!
Just a comment on an outdated dependency: I realize it is "just" the demo, but having an old base image will likely trigger vulnerability alerts pretty quickly after the first scan.

oid4vci/demo/frontend/Dockerfile
dbluhm: Good question; I think this may be influenced by the `create-react` template but I think we should be able to just bump this. @burdettadam?

burdettadam: Good catch! We should update the image.

dbluhm:
Node version in the demo dockerfile updated. I attempted to bump it to 20 (latest LTS) but was having some weird issues. We're going to be working on the demo soon so I just bumped to 18, which worked fine on the first try, and we'll revisit when we come back around.

esune (approved):
👍🏻

PR #48, opened 2023-12-02
Return redis_events tests

jamshale:
Bring back the redis_events integration tests that use additional plugin after repairing the build backend config line that was missing.

Also: changed the integration image names so faber using the integration dockerfile with the additional plugin is different from alice using the normal dockerfile in the docker directory.
Also: removed the `edit` condition for the PR workflow. It's not needed. 
Also: Had to add additional `$` character on the docker compose startup command on my linux machine. Not sure why? The workflow machine works with or without it.

### DIFF
--- a/.github/workflows/pr-integration-tests.yaml
+++ b/.github/workflows/pr-integration-tests.yaml
@@ -1,7 +1,7 @@
 name: Integration Tests
 on:
   pull_request:
-    types: [opened, edited, synchronize, reopened, ready_for_review]
+    types: [opened, synchronize, reopened, ready_for_review]
     branches:
       - "**"
   push:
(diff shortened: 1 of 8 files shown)


esune (approved):
Looks good to me, wondering about @WadeBarnes 's comment regarding the duplicate `Dockerfile` for the integration tests, but might be for another issue.

jamshale:
> Looks good to me, wondering about @WadeBarnes 's comment regarding the duplicate `Dockerfile` for the integration tests, but might be for another issue.

I couldn't think of a way to install the basic_message_storage plugin without another dockerfile. I agree it would be nice not to have an additional (almost duplicate) file.

Issue #28, opened 2023-11-02
Plug-In dependency managament and version mismatch issues


esune:
## The Problem

Each plug-in declares a `pyproject.toml` with a broad set of dependencies - including ACA-Py - and a lockfile generated at the time of the last code update. When the plug-in code is installed on an image (or other system) where those dependencies already exist, **poetry** will ensure the dependencies declared in the latest `pyproject.toml`/lockfile are respected and override previously installed dependencies altogether. This is true for all dependencies, but particularly bad for ACA-Py as it will cause the runtime to potentially have a version different than the one targeted (i.e.: different than what the official image tag declares) and cause issues.

## What to do?

While complex dependency management would be ideal, I also believe it is unlikely that we will want to create (and maintain) a dependency resolution system, especially since dedicated projects such as **pip** and **poetry** still seem to not have nailed this down.

What I would suggest is a mix of common sense and precautions when declaring dependencies: new dependencies in a plug-in should be kept at a minimum level, and dependencies that may conflict with the base system (such as ACA-Py) should be declared as dev dependencies rather than "regular" dependencies. This makes sense as plug-ins are not meant to be run standalone, and expect ACA-Py to be available on the system they are being installed on, and would prevent the installed version from being overridden at plug-in installation time.

This would still leave the issue for plug-ins that rely on a different (outdated or newer) version of ACA-Py to run: the solution in this case is simple: update the plug-in to run with the newer version of the dependency, or use an older image matching the requirement, whichever makes more sense for each use case.

A review of the currently available plug-ins and adjustments to each `pyproject.toml`, followed by the generation of new lockfiles would be the only required task to complete this.

esune:
Tagging @dbluhm @jamshale @usingtechnology @WadeBarnes @Jsyro @shaangill025 for input/feedback on this.

usingtechnology:
I like your thinking here. Since we are advocating the plugin be installed on an ACA-Py image then we can move the `aries-cloudagent` dependency from `main` to `dev` or its own group (similar to the integration dependencies).

dbluhm:
[...]

PR #49, opened 2023-12-04
Update ACA-Py dependencies

WadeBarnes:
Resolves: #28


redis_events/pyproject.toml
```diff
@@ -6,14 +6,20 @@ authors = ["Shaanjot Gill <gill.shaanjots@gmail.com>"]
...
 fastapi = "^0.68.0"
 nest-asyncio = "^1.5.5"
 pydantic = "^1.8.2"
 python = "^3.9"
 redis = "^4.1.4"
 uvicorn = "0.16.0"
 
+# Define ACA-Py as an optional/extra dependancy so it can be
+# explicitly installed with the plugin if desired.
+aries-cloudagent = { version = ">=0.10.3, < 1.0.0", optional = true }
```
jamshale, line 18: My repo-manager script, which I think will become more important as this repo grows, works off of empty lines between sections. I don't think it would handle the comments very well either because it sorts.

But I can change the logic a bit to handle things like this better. I think this approach makes sense. We'll have to do this for the other plugins as well but can merge this for now and then do the rest if this is successful.

jamshale:
I think this will work good. I'll just add a couple reviewers to get some other eyes on it.

dbluhm (approved):

WadeBarnes:
The tests on the `connection_update` plugin are failing due to an expired certificate on https://indy.igrant.io/

WadeBarnes:
The `basicmessage_storage` plugin is affected by the same issue.

WadeBarnes:
Ready to go.  Though we need to figure out what to do with the failing tests.  We either need to get `igrant.io` to update the certificate or switch to a different ledger like `http://test.bcovrin.vonx.io/`.  Open to suggestions.

swcurran:
Agree that using BCovrin Test is the right thing to do for this.

WadeBarnes:
Switching to `http://test.bcovrin.vonx.io/` fixed the issues with the failing integration tests.

multitenant_provider/integration/pyproject.toml
```diff
@@ -6,6 +6,7 @@ authors = []
 
 [tool.poetry.dependencies]
 aries-cloudagent = { version = "0.10.3" }
```
jamshale, line 8: Think this was missed

multitenant_provider/pyproject.toml
```diff
@@ -5,13 +5,20 @@ description = ""
...
-aries-cloudagent = { version = "0.10.3" }
 bcrypt = "^4.0.1"
 mergedeep = "^1.3.4"
 python = "^3.9"
 python-dateutil = "^2.8.2"
 typing-extensions = "4.5.0"
 
+# Define ACA-Py as an optional/extra dependancy so it can be
+# explicitly installed with the plugin if desired.
+aries-cloudagent = { version = "0.10.3" }
```
jamshale, line 16: here too

multitenant_provider/integration/pyproject.toml
WadeBarnes: The `multitenant_provider` didn't like it when it was updated to a newer version of ACA-Py, the tests failed, so I reverted it back to the version that worked.
See:
- https://github.com/hyperledger/aries-acapy-plugins/actions/runs/7090862684/job/19298755060
- https://github.com/hyperledger/aries-acapy-plugins/actions/runs/7090862685/job/19298755070

multitenant_provider/pyproject.toml
WadeBarnes: The `multitenant_provider` didn't like it when it was updated to a newer version of ACA-Py, the tests failed, so I reverted it back to the version that worked.
See:
- https://github.com/hyperledger/aries-acapy-plugins/actions/runs/7090862684/job/19298755060
- https://github.com/hyperledger/aries-acapy-plugins/actions/runs/7090862685/job/19298755070

multitenant_provider/integration/pyproject.toml
jamshale: OK. I can have a look after this is merged.

PR #58, opened 2023-12-14
[basicmessage_storage] Add message deletion and subwallet self disable. 

Jsyro:
https://github.com/bcgov/traction/issues/903

Add endpoint to delete message by message id. 

Add setting `basicmessage_storage.wallet_enabled` that allows subwallets to disable message storage. 

Has unit tests, but integration tests are not setup for multi-tenancy, can be added later if required.


jamshale (approved):
LGTM

Jsyro:
> Looks good to me. @Jsyro would there be a way to set the default storage behaviour to `False` (not store messages) and have each tenant opt-in, rather than the other way around?

Easy change, do we want it to always be false? or is should there be a top level config that is inherited by default?

This plugin does one thing, so I don't want have someone install it to miss there is also a config to enable.

esune:
> > Looks good to me. @Jsyro would there be a way to set the default storage behaviour to `False` (not store messages) and have each tenant opt-in, rather than the other way around?
> 
> Easy change, do we want it to always be false? or is should there be a top level config that is inherited by default?
> 
> This plugin does one thing, so I don't want have someone install it to miss there is also a config to enable.

Top-level config seems like a good idea to me, it follow the patterns we have in place with startup parameters and other settings. The global plugin config would define the default behaviour, and then each tenant can override as-needed.

esune (approved):

Jsyro:
Before I merge, this is a breaking change (without configuration changes to existing agents, their behaviour will change), are there corresponding instances/configurations that need to be updated?

esune:
I don't think we are (yet) using the plugin code from this repo in Traction, and I am not aware of anybody else who would be using it either.

PR #64, opened 2023-12-30
Update repo manager

jamshale:
This PR ended up being larger than I thought.

Script:
- Updated the script to handle spaces, comments and extra sections. Made small adjustment to oid4vci libraries. Upgraded askar to ~3.0 globally. Working really well.
- Ran the script to update dependencies.
- Removed the common development files part of the script. I thought they would be the same for every plugin but I'm already seeing them be customized in @Jsyro latest PR.
- Made the script smaller and easier to maintain. Still a bit over work could be done to use the same code for the main vs integration test sections. 

After upgrading all the plugins I had to fix some tests. 
- Moved the oid4vci unit tests to the inner project
- Fixed a couple unit tests in multitenant_provider (not sure why these broke)
- Fixed kafka_events integration tests after pytest upgrade.
- Fixed multitenant_provider integration tests. There is currently an issue with the log file when agent run in multitenant admin mode.
- Fixed redis events integration tests. Using basic_message storage which now requires it to be enabled.
- changed test result files to be generated in the plugins so script can ignore that directory and just better to have them isolated.

All plugins are now using the same global packages.


dbluhm (approved):
LGTM!

PR #78, opened 2024-01-26
feat: DIDComm RPC protocol

amanji:
This PR adds support for the DIDComm RPC (DRPC) protocol for ACA-Py as a plugin. The protocol has been implemented largely according to [Aries RFC 0804](https://github.com/hyperledger/aries-rfcs/blob/bfa0bb23bda457a92a2a37a33b396f43ae84eb03/features/0804-didcomm-rpc/README.md) with internal objects that  conform to the  [JSON RPC specification](https://www.jsonrpc.org/specification).

Minor modifications to Aries RFC 0804:
* "All JSON-RPC requests are notifications": It was not specified how responses to these should be handled. Since this protocol does not make use of Acknowledgements or Problem Reports, I have adapted how JSON-RPC handles batch requests and require that an empty object or array be returned by the responding Agent in order for the interaction to be considered complete.

As of now there is nothing that I am aware of that is incomplete, except for the question of whether additional validation should be included to check that IDs in a JSON-RPC response correspond to the IDs received in a JSON-RPC request. Since it is expected that controllers understand exactly how to interpret JSON-RPC request and response/error objects, this may be superfluous.

This implementation includes a suite of unit and integration tests that meet the suggested requirement of at least 90% coverage.


amanji:
Tagging @ianco and @dbluhm

jamshale:
For the failing integration tests I think you just need to run `poetry install` once from inside the `integration` folder. This should be build into the initialization script at some point.

.devcontainer/Dockerfile
```diff
@@ -0,0 +1,45 @@
+FROM mcr.microsoft.com/devcontainers/base:jammy
```
jamshale, line 1: Maybe this folder got generated automatically by mistake? I would like to avoid having a devcontainer config for the root of this project.

amanji: I might have generated it myself so I could run the plugin in a devcontainer of its own

rpc/rpc/v1_0/handlers.py
```diff
@@ -0,0 +1,84 @@
...
+                type=DRPCRecord.RECORD_TYPE,
+                value=json.dumps(request_record.serialize()),
+                tags={
+                    "connection_id": connection_id,
+                    "thread_id": thread_id,
+                },
+            )
+            await storage.add_record(record)
+
+        await context.profile.notify(
```
jamshale, line 45: Is there a event bus handler for this? I'm not seeing where it's registered. Is this doing anything or supposed to be?

I don't have it open in my code editor so might just be missing something

amanji: There isn't. I wasn't sure where to register the events

jamshale:
I think overall looking really good :+1: Will try and go over things more low level at some point.

rpc/rpc/v1_0/handlers.py
jamshale: If you want to register the events you can do it like this in your routes.py file.

image

The naming is important. The agent looks in all the routes.py files for this method when intializing the event bus. You can see them getting registered with log-level set to debug.
 
If it doesn't need to do anything with this message than you don't need this though.

amanji: I definitely want controllers to be notified of incoming messages for this protocol so this helps.

.devcontainer/Dockerfile
amanji: Is there a best practice for this? I have a devcontainer config at the root of the plugins repo in order to generate an environment to run the `repo_manager` script. I also have a devcontainer config in the plugin directory to generate an environment specifically for plugin development.

jamshale: You only need python and poetry installed on your computer to run the script. I guess we could have a dev container for that but I'd rather not. The plugin environment is a lot more complicated and benefits more from the dev container env.

amanji: I can remove that root config then

rpc/rpc/v1_0/handlers.py
amanji: Will the event bus result in web hook callbacks? That's really what I'm looking to accomplish

jamshale: Oh. I'm not an expert on this but I don't think so. The event bus is for messages within one agent. Which is handy for decoupling components.
Do these agents have a connection?
If you want to notify another agent then I think you need to use the `BaseResponder --> aries_cloudagent/messaging/responder.py`. You can inject it and then notify another agents handler.

Ian or Daniel might be able to help better though. I haven't done a lot of stuff in this area.

amanji: This would be in the case of an Agent notifying its own controller(s) through the webhooks. Perhaps @ianco or @dbluhm could weigh in.

dbluhm: The webhook emitter is itself an event listener listening on a specific set of events:

https://github.com/hyperledger/aries-cloudagent-python/blob/main/aries_cloudagent/admin/server.py#L47-L59

amanji: So that means I'll have to make a corresponding change in ACA-Py directly, since plugins up to this point haven't had the requirement of calling back to controllers. Is that correct? Do we want to generalize this in the plugins or should I just add this to the current list of events in ACA-Py for now?

amanji (commented):
Need to update PR description and README

rpc/rpc/v1_0/handlers.py
dbluhm: I think what I would recommend is to emit two events: 
- One for `drpc::request::received`; this could be consumed by plugin based rpc handlers
- One for `acapy::webhook::drpc-request`; this would be consumed by the admin server webhook listener to then in turn be consumed by controller based rpc handlers

amanji: Ok that makes sense. I'll emit a second event which we'll register with the Admin event listeners

dbluhm: The admin server is already listening for `acapy::webhook::(.*)`, fortunately, so no changes should be required upstream. It will turn it into a webhook to `/topic/drpc-request` (maybe we should do an underscore instead of a hyphen since I think that's what is used in other webhook topics).

amanji: Tbh we don't really need the first event emitted. The DRPC handlers don't listen for those events.

dbluhm: No events inside of this plugin, sure. But I like the idea of being able to define handlers from another plugin so I think the first event would be handy to keep around. And listening on `acapy::webhook::drpc-request` is doable but feels odd. But I don't feel strongly on that point.

amanji: I meant to have that in snake case. I can add the second emitter back as well.

jamshale (commented):
Looking good. Will wait to review better until you open it but I can't see any issues.

jamshale (approved):
I'm approving this. It looks really good to me. It needs documentation but I'm fine with that being added later.

jamshale:
Next time, a overview as the PR description would be good. I still don't know a lot about the use case for this, but I trust you do, and the code/testing looks good.

This repo squashes commits so you don't need to do anything there.

ianco:
> I'm approving this. It looks really good to me. It needs documentation but I'm fine with that being added later.

We need some documentation with this PR, even if it's minimal.  We too often leave documentation for later and then never provide it.

ianco (changes requested):
Documentation

rpc/rpc/v1_0/routes.py
```diff
@@ -0,0 +1,255 @@
...
+            return web.json_response(msg.serialize())
+
+        raise web.HTTPForbidden(reason=f"Connection {connection_id} is not ready")
+
+
+async def register(app: web.Application):
+    """Register routes."""
+
+    app.add_routes([web.post("/drpc/request", drpc_send_request)])
+    app.add_routes([web.post("/drpc/response", drpc_send_response)])
```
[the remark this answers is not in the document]
amanji: I added some follow-up commits that add some endpoints to fetch records

swcurran:
@amanji — do you have the minimal documentation to get this published?  When do you think you can get this done so we can merge this?

amanji:
Just tested with Traction. Documentation will be completed today

jamshale (approved):

ianco (approved):

Issue #84, opened 2024-02-01
Devcontainers should include Poetry version 1.6.1 (or 1.7.1?)


loneil:
The ACA-Py libraries are built with 1.6.1
https://github.com/hyperledger/aries-cloudagent-python/blob/main/poetry.lock

so probably want to stay in sync for this repo's Poetry usage in the devcontainer? Currently `"POETRY_VERSION": "1.4.2"`
https://github.com/hyperledger/aries-acapy-plugins/tree/main/plugin_globals/.devcontainer

loneil:
Each individual plugin seems to have been generated on 1.7.1 though. But the devcontainer definitions say 1.4.2 as well

jamshale:
I ran the script locally that's why it's 1.7.1. Dev containers should be updated :heavy_check_mark:

loneil:
Ok thanks. I did a PR to update them here. Found 1.4.2 references in dockerfiles too
https://github.com/hyperledger/aries-acapy-plugins/pull/85

PR #85, opened 2024-02-02
Normalize poetry versions in devcontainers/docker to match lockfiles

loneil:
The `poetry.lock` files have been generated by 1.7.1 for the plugins, but looks like the devcontainers and dokerfiles reference Poetry 1.4.2.

This PR sets them all to the matching 1.7.1

ACA-Py actually uses 1.6.1 so maybe that would be more appropriate, but ACA-Py should probably update to a newer version with 0.12.0 (?) or a subsequent release. Also it probably doesn't matter to have the slight mismatch between these plugins and ACA-Py for Poetry version
https://github.com/hyperledger/aries-cloudagent-python/blob/main/poetry.lock

### DIFF
--- a/basicmessage_storage/.devcontainer/Dockerfile
+++ b/basicmessage_storage/.devcontainer/Dockerfile
@@ -2,7 +2,7 @@
 ARG VARIANT="3.9-bullseye"
 FROM mcr.microsoft.com/vscode/devcontainers/python:0-${VARIANT}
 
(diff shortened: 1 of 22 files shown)


loneil:
Issue https://github.com/hyperledger/aries-acapy-plugins/issues/84

jamshale (approved):

PR #86, opened 2024-02-02
Fix outbound queue with webhook-url

jamshale:
This fixes a bug where an empty header value was added to the outbound payload which caused the outbound transit to fail. Also, ignore webhook forwarding for messages which don't match the webhook event map.

Added a available dummy server to the integration test network and enabled the webhook url config in the integration tests.

I still don't understand why pushing an invalid outbound message into the queue is actually blocking/breaking the queue from continuing to work. This is a problem, as this small one line mistake, caused the invalid payload. I'll be trying to figure out how to prevent this, but this is still a fix/improvement for the current problem. 

@jpnortherblock

### DIFF
--- a/redis_events/redis_events/v1_0/redis_queue/events/__init__.py
+++ b/redis_events/redis_events/v1_0/redis_queue/events/__init__.py
@@ -1,21 +1,21 @@
 """ACA-Py Event to Redis."""
 
+import base64
 import json
 import logging
-import base64
 import re
 from string import Template
 from typing import Any, Optional, cast
 
+from aries_cloudagent.config.injection_context import InjectionContext
 from aries_cloudagent.core.event_bus import Event, EventBus, EventWithMetadata
 from aries_cloudagent.core.profile import Profile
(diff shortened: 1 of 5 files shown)


redis_events/redis_events/v1_0/redis_queue/events/__init__.py
```diff
@@ -130,35 +128,30 @@ async def handle_event(profile: Profile, event: EventWithMetadata):
...
                 api_key = None
                 if len(endpoint.split("#")) > 1:
                     endpoint_hash_split = endpoint.split("#")
                     endpoint = endpoint_hash_split[0]
                     api_key = endpoint_hash_split[1]
-                webhook_topic = config_events.event_webhook_topic_maps.get(
-                    event.topic)
-                endpoint = f"{endpoint}/topic/{webhook_topic}/"
-                headers = {"x-wallet-id": wallet_id} if wallet_id else {}
-                if not api_key:
```
jamshale, line 146: This was the mistake...

dbluhm (approved):

PR #99, opened 2024-02-06
Bugfix: vc.credentialSubject.id, vc.issuer, iss, and sub should be DID itself

t-kanuma:
[OID4VCI]
credentialSubject.id, iss, and sub should be DID, not DID URLs pointing to verification methods.

### DIFF
--- a/oid4vci/oid4vci/public_routes.py
+++ b/oid4vci/oid4vci/public_routes.py
@@ -350,25 +350,31 @@ async def issue_cred(request: web.Request):
     if not pop.holder_kid:
         raise web.HTTPBadRequest(reason="No kid in proof; required for jwt_vc_json")
 
+    verification_method = ex_record.verification_method
+    issuer_did = verification_method.split("#")[0] if verification_method.startswith("did:") else verification_method
+
+    holder_kid = pop.holder_kid
+    holder_did = holder_kid.split("#")[0] if holder_kid.startswith("did:") else holder_kid
+
     payload = {
         "vc": {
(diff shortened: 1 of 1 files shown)


dbluhm:
Could you point me to relevant documentation on this? Thanks!

t-kanuma:
@dbluhm 
um, probably here?

https://www.w3.org/TR/vc-data-model/#identifiers
> The value of the id [property](https://www.w3.org/TR/vc-data-model/#dfn-property) MUST be a single [URI](https://www.w3.org/TR/vc-data-model/#dfn-uri). It is RECOMMENDED that the [URI](https://www.w3.org/TR/vc-data-model/#dfn-uri) in the id be one which, **if dereferenced, results in a document containing machine-readable information about the id.**

I'm a beginner in this field, so I'm not quite sure, but I've never seen before that 'credentialSubject.id' and 'issuer' values are DID URLs pointing to Verification Methods, and it feels uncommon to me. Is there any rational reason for this?

dbluhm:
Thanks! And welcome to the field! I think you must be right. I'm trying to think back to why we chose to use DID URLs to verification method IDs at the time. I might have to spend some time looking deeper at this. Unfortunately, that might be a bit before I can look too much closer.

Would you be able to look into the test failures? DCO especially will need to be taken care of. The `Details` link should provide some pretty detailed instructions on how to handle that. Thanks!

t-kanuma:
Thanks.

> Would you be able to look into the test failures?
Yes sure. I'll take care of it soon.

PR #103, opened 2024-02-09
Upgrade fastapi

jamshale:
Upgrade fastapi for security issues


dbluhm (approved):

jamshale:
> Will a similar change be necessary in Kafka events? I think the deliverer API is very similar.

I think so... I thought I saw something flagged in security for it but dependabot hasn't created PR's for it. I did a PR for this because the integration tests failed but I think it was non related to the fastapi upgrade.

PR #118, opened 2024-02-21
:construction_worker: add pip to dependabot management

ff137:
After workflow actions have been enabled, just to get builds to run and auto dependabot job to start after merge


ff137:
Wrong base repo ...

Issue #37, opened 2023-11-15
Tagging this repo


dbluhm:
The concept of tagging this repo came up on the Aries WG call on Nov. 15, 2023. I wanted to continue that discussion here:

As prior art, consider this text from: https://github.com/hyperledger/aries-acapy-plugin-toolbox#aca-py-version-compatibility

> ## ACA-Py Version Compatibility
> 
> To avoid a confusing pseudo-lock-step release, this plugin is versioned independent of ACA-Py. Plugin releases will follow standard [semver](semver.org) but each release will also be tagged with a mapping to an ACA-Py version with the format `acapy-X.Y.Z-J` where `X.Y.Z` corresponds to the ACA-Py version supported and `J` is an incrementing number for each new plugin release that targets the same version of ACA-Py.
> 
> You should look for the most recent release tagged with the version of ACA-Py you are using (with the highest value for `J`).

This was a convention that the toolbox plugin adopted (back when we were actively working on it). I think tagging the whole repo will be challenging to do meaningfully; for the toolbox plugin, we needed to keep track changes of the plugin itself as well as the versions of ACA-Py it was compatible with. The plugins should be able to evolve and gain features and be able to version themselves to communicate when new features are added.

jamshale:
The plugins poetry `pyproject.toml` file is kind of doing this already. It was the aries-cloudagent and python version that the plugin has been tested. How to use it when trying to pull a version of the plugin though I'm not sure.

I had been thinking that we would keep all the plugins up to date with acapy and that each tag would have release notes with what version of acapy and python it was tested on. This might not be maintainable though. Currently all agents are on the same acapy version 0.11.0. 

Not sure what the best thing to do would be but would really like to get this completed and have a reliable way to reference frozen plugins.

jamshale:
This has been completed.

PR #119, opened 2024-02-21
:construction_worker: Configure dependabot to manage pip dependencies for each plugin

ff137:
I notice that the dependabot file only manages github-actions

It makes sense to start defining pip dependency management for each plugin.

This PR just adds daily interval scheduling for each plugin.

This _will_ spam with a lot of new dependabot PRs initially, but I think that's good to get out of the way and start having an automated workflow for dependency management.

___

_**Note**_: this may increase the urgency of more frequent tagged releases. Currently this repo has no tagged releases. This means developers installing from this repo can only point to main, commit hashes, or will have to manage their own fork. 

No tagged releases makes version control harder, and developers pointing to main may get problems if they run an older ACA-Py instance which is not in the automated test workflows. e.g. if the tests run on ACA-Py 0.11, then tests may pass when it causes a dependency conflict in ACA-Py 0.10.

So, if this new dependabot workflow is merged, together with the future dependabot updates, it should probably come with an improved release cycle.

I propose daily releases ... when there are new changes. Which can probably be automated to just publish a tag with that day's date. That way there's better version control for people using these plugins. If something breaks for those pointing to main, they can easily point to an older tag. Not sure what alternative is best, considering each individual plugin will have its own development and update cycle.

These PRs can also be scheduled to come weekly, and then there are weekly releases?

Let me know what you think!

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,5 +1,5 @@
- # For details on how this file works refer to:
- #   - https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file
+# For details on how this file works refer to:
+#   - https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file
 version: 2
 updates:
   # Maintain dependencies for GitHub Actions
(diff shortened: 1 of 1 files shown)


jamshale (approved):

jamshale:
Thanks. Looks good!

ff137:
@jamshale cool! Please remember to discuss the question around more frequent releases with the team. I'm just cognisant that upgrading dependencies can break stuff for people pointing to main. Maybe not likely, but still! Daily or weekly tagged releases will help with that

jamshale:
> @jamshale cool! Please remember to discuss the question around more frequent releases with the team. I'm just cognisant that upgrading dependencies can break stuff for people pointing to main. Maybe not likely, but still! Daily or weekly tagged releases will help with that

I very much agree. The problem is we haven't decided on a release strategy. Mainly, if and how we want to tag releases in a way that relates to acapy. https://github.com/hyperledger/aries-acapy-plugins/issues/37. I'm going to try and put more thought into this and have tagged releases for this project. Any thoughts are very welcome.

Currently we are basically hoping the integration and unit testing is enough to catch breaking changes and bugs when updating libraries, but having known stable releases is definitely something we want to have.

PR #228, opened 2024-03-05
Upgrades

jamshale:
General upgrades. Increased some specification on dev dependencies. Fixed some warnings in kafka events.


dbluhm (approved):

ff137:
Just a note: the `redis_events/pyproject.toml` file didn't get the same love with the upgrades!
The poetry.lock was updated but not the toml


Mentioning it because we are recently making use of the redis_events plugin. And I'm curious if some of the upgrades break anything in the integration, or if it was missed.

jamshale:
> Just a note: the `redis_events/pyproject.toml` file didn't get the same love with the upgrades! The poetry.lock was updated but not the toml
> 
> Mentioning it because we are recently making use of the redis_events plugin. And I'm curious if some of the upgrades break anything in the integration, or if it was missed.

I did have an issue with the redis_events plugin in particular. So I installed from the devcontainer which didn't get the global updates. The script was trying to upgrade aries-cloudagent to 0.12.0-rc1which I didn't want. I think it might be a caching issue or something because all the other plugins were correctly on 0.11.0.

I'll try and figure out why that was happening but there wasn't any problems with those upgrades that I know of.

PR #387, opened 2024-04-22
Create Automated Release Workflows

jamshale:
There's some fairly complicated logic in a few places so I created a couple flow charts at `ReleaseWorkflows.md`.

Basically, this is a way to keep the plugins up to date with acapy releases automatically. If all the tests pass for a plugin it will be updated and be considered stable with acapy. If it doesn't then it won't be upgraded and will remain verified against an earlier release. A maintainer can create patch releases if a plugin which did fail for a release gets fixed before another version of acapy. A message is appended to the plugin description as well with the supported acapy version. I'm hoping this can be used to provide warnings in acapy when a plugin falls behind acapy.

There is some bash scripting. I tried to limit it as much as I could by utilizing the `repo_manager.py` script. The manager script has been changed slightly and some additional options for creating and parsing release information has added.

I had to do a bit of work for some tests after the upgrade but they all passing now so they will all be on version 0.12.0 of acapy initially. Two unit tests for rpc I removed. I couldn't figure out why they were only failing in the github action but not locally.

*** Note: *** I'm not sure if the bot will be able to create the release PR's for this repo. Might need to change it to a fork strategy.


.github/workflows/create-release-pr.yaml
```diff
@@ -0,0 +1,328 @@
...
+          echo "Input re-release = ${{ inputs.re_release }}"
+          if [ ${{ inputs.re_release }} == "true" ]; then
+            echo "Re-Release requested"
+          else
+            if [ $(sem_version $current_global_version) -ge $(sem_version $current_available_version) ]; then
+              echo "Version of aries-cloudagent is up to date"
+              exit 1
+            fi
+              echo "Detected aries-cloudagent upgrade available..."
+          fi
```
WadeBarnes, line 95: Would it be better to use conditional logic on the step rather than checking `inputs.re_release` inside the script?

.github/workflows/create-release-pr.yaml
```diff
@@ -0,0 +1,328 @@
...
+
+          # Create the PR
+
+          details="Plugins upgraded this release: "
+          for plugin in ${upgraded_plugins}; do
+            details="$details $plugin"
+          done
+
+          title="Release for aries-cloudagent v$release_version"
+          gh pr create --title "$title" --body "$body $details" --base main --head "release-v$release_version"
```
WadeBarnes, line 328: Few things here:
- Would it be better to use conditional logic on the step rather than checking `steps.integration_test_plugins.outputs.integration_test_exit_code` inside the script?
- `GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}` environemnt variable is not used.  It's better to use `${{ secrets.GITHUB_TOKEN }}` directly anyway.
- There's a better way to update code and submit a PR.  See https://github.com/hyperledger/indy-node/blob/main/.github/workflows/tag.yaml

.github/workflows/create-release.yaml
```diff
@@ -0,0 +1,119 @@
...
+          details="Plugins upgraded this release: "
+          upgraded_plugins=$(python repo_manager.py 5)
+          for plugin in ${upgraded_plugins}; do
+            details="$details $plugin"
+          done
+
+          git tag -a $release_tag -m "Release $tag"
+          git push origin $release_tag
+          gh release create $release_tag --title "Release for aries-cloudagent v$version" --notes "$body $details"
+
```
WadeBarnes, line 119: There is a better way to create releases.  See https://github.com/hyperledger/indy-node/blob/main/.github/workflows/publishRelease.yaml

WadeBarnes (changes requested):
@jamshale, A few suggestions below.

WadeBarnes:
@jamshale, One question.  It looks like a release is done for the plugins as a whole.  Is that correct?  Is the plan to later support releasing the plugins individually so they can evolve independently?

jamshale:
> @jamshale, One question. It looks like a release is done for the plugins as a whole. Is that correct? Is the plan to later support releasing the plugins individually so they can evolve independently?

With this PR, yes, you are correct. The entire repo is released at the same time as acapy. However, it does allow for plugins to be upgraded independently. So plugins can now be left in a non-upgraded state indefinitely, and only say they are fully supported for an older version of aca-py. When installing a specific plugin the aca-py agent could target any of the release tags and the subdirectory of the plugin.

There is no publish step yet. I was thinking that this could be done as a separate task. When publishing to pypi only plugins which had been upgraded would be published. Allowing them to evolve independently as a published package.

I didn't have a solid plan about having a common naming or package versioning. I wasn't sure if we wanted plugins versions to match aca-py or not, or have there own versioning. Currently most of them are just `0.1.0` and don't get changed with this workflow. The plugin names don't get `aries-plugin` or anything pre-pended to the package name either.

Either of these things would be easy to do. All I do in this workflow is add the supported aca-py version in the package description.

jamshale:
Please ignore any WIP commits happening here when I work on the improvements.

jamshale:
Going to close a re-open this later. Need to figure some stuff out that requires pushing a lot of commits.

PR #391, opened 2024-04-24
Create Automated Release Workflows 

jamshale:
There's some fairly complicated logic in a few places so I created a couple flow charts at `ReleaseWorkflows.md`.

Basically, this is a way to keep the plugins up to date with acapy releases automatically. If all the tests pass for a plugin it will be updated and be considered stable with acapy. If it doesn't then it won't be upgraded and will remain verified against an earlier release. A maintainer can create patch releases if a plugin which did fail for a release gets fixed before another version of acapy. A message is appended to the plugin description as well with the supported acapy version. I'm hoping this can be used to provide warnings in acapy when a plugin falls behind acapy.

There is some bash scripting. I tried to limit it as much as I could by utilizing the `repo_manager.py` script. The manager script has been changed slightly and some additional options for creating and parsing release information has added.

I had to do a bit of work for some tests after the upgrade but they all passing now so they will all be on version 0.12.0 of acapy initially. Two unit tests for rpc I removed. I couldn't figure out why they were only failing in the github action but not locally.

*** Note: *** I'm not sure if the bot will be able to create the release PR's for this repo. I have tried to use the org secret and replicated the same thing for my fork, so hopefully it does.


.github/workflows/create-release-pr.yaml
```diff
@@ -0,0 +1,357 @@
...
+          git add ./RELEASES.md
+
+          git commit -s -m "Release v$release_version Upgrades"
+
+          # Prepare the PR body
+
+          EOF=$(dd if=/dev/urandom bs=15 count=1 status=none | base64)
+          echo "pr_body<<$EOF" >> $GITHUB_OUTPUT
+          echo "$body $details" >> $GITHUB_OUTPUT
+          echo "$EOF" >> $GITHUB_OUTPUT
```
WadeBarnes, line 340: I don't think the `git` operations are necessary here.  From what I understand that will all be handled by the `peter-evans/create-pull-request` in the next step.

.github/workflows/create-release.yaml
```diff
@@ -0,0 +1,137 @@
...
+            fi
+          done
+
+          if [ "$found_upgrade" = true ]
+          then
+            echo "Upgrade Detected. Creating Release"
+          else
+            echo "No Upgrade Detected. Skipping Release Creation."
+          fi
+          echo should_create_release=$found_upgrade >> $GITHUB_OUTPUT
```
WadeBarnes, line 68: File filters could be used to ensure, overall, the workflow is only triggered when `lock` files have been updated.

.github/workflows/create-release.yaml
```diff
@@ -0,0 +1,137 @@
...
+          done
+
+          # Set the outputs
+
+          echo tag=$release_tag >> $GITHUB_OUTPUT
+
+          EOF=$(dd if=/dev/urandom bs=15 count=1 status=none | base64)
+          echo "release_body<<$EOF" >> $GITHUB_OUTPUT
+          echo "$body $details" >> $GITHUB_OUTPUT
+          echo "$EOF" >> $GITHUB_OUTPUT
```
WadeBarnes, line 123: Can you provide an example of the resulting tagging format in ReleaseWorkflows document?

WadeBarnes (changes requested):

.github/workflows/create-release-pr.yaml
jamshale: I think you may be right. I may be able to remove some of them. I was following this section https://github.com/peter-evans/create-pull-request?tab=readme-ov-file#create-your-own-commits in the readme.

The thing was I am using git to remove the changes for the failed upgrades. I may be able to remove the add and commit statements at the end though.

PR #418, opened 2024-05-02
Release workflows

jamshale:
Opening this as a new PR again. I had to do work on my forked main branch and when I rebased it forced closed the old PR's.  
https://github.com/hyperledger/aries-acapy-plugins/pull/391
https://github.com/hyperledger/aries-acapy-plugins/pull/387

I think I addressed all the comments. 

Some of the bash scripting is pretty gross but I tried to comment and document it well.

I had to fix some tests and I upgraded the anoncreds library as well. This is where the changes unrelated to the workflows came from.

### DIFF
--- a/oid4vci/integration/Dockerfile
+++ b/oid4vci/integration/Dockerfile
@@ -1,4 +1,4 @@
-FROM python:3.10-slim-bullseye AS base
+FROM python:3.10-slim-bookworm AS base
 WORKDIR /usr/src/app
 
 ENV POETRY_VERSION=1.5.0


oid4vci/integration/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM python:3.9-slim-bookworm AS base
+FROM python:3.10-slim-bullseye AS base
```
WadeBarnes, line 1: `bookworm` => `bullseye`, is a downgrade from Debian 12 to Debian 11.  Is this intentional?

WadeBarnes (approved):
Other than the one question below, the workflows look good from a high level.  I did not review the code in the various steps in detail.

oid4vci/integration/Dockerfile
jamshale: No. Not intentional. Just needs to be 3.10. Updated.

PR #459, opened 2024-05-14
update release PR workflow token

jamshale:
I tried using the new PAT token for the PR workflow but it didn't work. I'm thinking it possibly has the permissions required to make the release but not the PR. The normal GITHUB_TOKEN should have the permissions to make the PR so i'm trying that.

Screenshot from 2024-05-14 08-29-27

### DIFF
--- a/.github/workflows/create-release-pr.yaml
+++ b/.github/workflows/create-release-pr.yaml
@@ -360,7 +360,7 @@ jobs:
         with:
           author: Release Bot <release-bot@users.noreply.github.com>
           committer: Release Bot <release-bot@users.noreply.github.com>
-          token: ${{ secrets.BOT_PR_PAT }}
+          token: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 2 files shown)


WadeBarnes (changes requested):
Remove `permissions` from the global scope; lines 15 and 16.

Narrow the `permissions` on the job level, `write-all` is too permissive.  According to the docs for [peter-evans/create-pull-request](https://github.com/peter-evans/create-pull-request?tab=readme-ov-file#action-inputs), you will only need `contents: write` and `pull-requests: write`.

jamshale:
Changed the permissions. I'm actually not certain anymore if the PAT token was necessary :/ I thought I had to do it, but I might have been wrong, and had them wrong in my forked repo. I guess it's not a big deal if it has similar permissions as the GITHUB_TOKEN.

WadeBarnes (approved):

PR #460, opened 2024-05-15
OID4VCI plugin - Fix delete  exchange-supported records

timbl-ont:
Quick fix on ./oid4vci/exchange-supported/records delete. There is a typo that duplicates the "credential supported identifier" as a parameter causing deletes to fail.

FYI @dbluhm

### DIFF
--- a/oid4vci/oid4vci/public_routes.py
+++ b/oid4vci/oid4vci/public_routes.py
@@ -350,7 +350,7 @@ async def issue_cred(request: web.Request):
     if not pop.holder_kid:
         raise web.HTTPBadRequest(reason="No kid in proof; required for jwt_vc_json")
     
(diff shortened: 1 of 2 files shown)


dbluhm (approved):

jamshale:
You'll need to sign your commits. Then we can merge this.

timbl-ont:
@jamshale DCO has been added.

timbl-ont:
@jamshale thanks for the feedback. It is fixed now. I am switching between two github accounts and did not realize my DCO remediation PR was with a different github config from the initial PR. Anyway, fixed now after signing off on all the authors.

jamshale:
Dang! there's just a linting issue now at `oid4vci/public_routes.py:353:91:` . You can just add `#E501`to the end of the line.

This ones a bit annoying but it's the same in aca-py.

PR #494, opened 2024-05-21
chore: updating dependabot to support GHA, python, JS, TS, docker and dev container packages

rajpalc7:


rajpalc7:
Hi @WadeBarnes, this is ready for review now.

WadeBarnes (approved):
LGTM, but I'll leave the final review and approval to the regular maintainers.

jamshale:
I'm not sure if we want to manage adding another section to the dependabot config everytime. I didn't know it was like this already. Although the changes to weekly are good.

Is it possible just to have a single config for each type at the root directory?

WadeBarnes:
Unfortunately you have to point Dependabot at the folder containing the dependency manifests.  It's not smart enough to search the folders to find the manifests.

swcurran:
Having instructions to update the dependabot config when adding a plugin is not a bad thing. We don't add them that often...

I assume that is the issue?

jamshale:
> Having instructions to update the dependabot config when adding a plugin is not a bad thing. We don't add them that often...
> 
> I assume that is the issue?

Ya. Mostly just managing all the places. Some of the plugins have multiple places/configs nested inside them. For the simpler plugins it's only the root and integration test folder.

We can merge it though and see how it goes. Maybe figure something out in the future.

PR #567, opened 2024-05-28
Release 1.0.0

jamshale:
I had to do the release pr manually this time because of the python upgrade. When pulling the aries-cloudagent version with 3.9 it won't get the 1.0.0 release, so it doesn't run the workflow.

Does all the required changes to the source code. Mostly around the authentication decorators and pydantic upgrade.

Removes black from most dependencies and the bbs libraries.

kafka_events is troublesome for using python 3.12 with the aiokafka library in `alpine`. I reverted it back to version 3.9 for now. This is the only plugin that can't be used with version 1.0.0 currently (https://github.com/hyperledger/aries-acapy-plugins/issues/746)


jamshale:
Sounds good. I'll merge that and we can test the minor patch release with kafka events if it becomes available.

PR #768, opened 2024-07-15
Fixes and interop testing

dbluhm:
This PR adds interop testing against Credo and Sphereon's OID4VCI client library. There were also some fixes to achieve interop:

- Issuer ID recorded in OID4VCI exchange record.
- Fill in the `iss` and `credentialSubject.issuer` fields using record Issuer ID instead of the verification method ID.

Heads up @timbl-ont @weiiv, there might be some minor conflicts with the work you've been doing.


oid4vci/integration/credo/index.ts
```diff
@@ -0,0 +1,168 @@
...
+              jwk: getJwkFromKey(key),
+            }
+          }
+
+          throw new Error('Unable to create a key binding')
+        },
+      }
+    )
+
+    console.log('Received credentials', JSON.stringify(credentials, null, 2))
```
jamshale, line 150: Is this intentional?

oid4vci/integration/sphereon/index.ts
```diff
@@ -0,0 +1,83 @@
...
+      .setIssuedAt()
+      .setExpirationTime('2h')
+    if (kid) {
+      jwt.setIssuer(kid)
+    }
+    if (args.payload.aud) {
+      jwt.setAudience(args.payload.aud)
+    }
+    console.log('signing: ', jwt)
+    console.log(privateKey)
```
jamshale, line 63: Intentional?

oid4vci/pyproject.toml
```diff
@@ -31,7 +31,7 @@ pytest-ruff = "^0.3.2"
 aries-askar = { version = "~0.3.0" }
 indy-credx = { version = "~1.1.1" }
 indy-vdr = { version = "~0.4.1" }
-ursa-bbs-signatures = { version = "~1.0.1" }
```
jamshale, line 34: I've never known when this is required and when not? I think it's installed for every plugin, so we should probably change that.

jamshale (commented):
I will approve. Just wasn't sure if some of the console logs should be there or not, or if they were used for debugging?

oid4vci/integration/credo/index.ts
dbluhm: Yeah, handy for debugging; since this is meant purely for testing, I think it's valuable to keep around.

oid4vci/integration/sphereon/index.ts
dbluhm: Ditto

oid4vci/pyproject.toml
mepeltier: That may have been a stray change from me—that package causes issues on Apple Silicon, so I just removed it when I was testing, since this plugin doesn't use it. I do think you're right, in that it's installed for every plugin, so reverting that might not be a bad idea for consistency if nothing else

dbluhm: Inclusion was making it difficult for @mepeltier to run the plugin since he's on M* Mac. I'll have a think about what the right way to handle this is :thinking:

dbluhm: Looking over this again, I'm actually a bit confused what the integration dependency group represents. At least in this plugin, we have a separate poetry setup in the integration test directory that handles integration test dependencies.

jamshale: I think the idea was you have the required deps to run the agent and plugin in the outer level and then have additional ones in the integration test directory. I agree it should maybe be rethought a bit.

jamshale: > Inclusion was making it difficult for @mepeltier to run the plugin since he's on M* Mac. I'll have a think about what the right way to handle this is 🤔

Maybe just don't commit this change for now? Can leave it commented when developing on mac.

jamshale (approved):

oid4vci/pyproject.toml
dbluhm: Reverted

PR #774, opened 2024-07-19
Allow for `lite` plugins

jamshale:
Allow for `lite` plugins to be listed in a text file and these will be skipped by the integration test management script and CI/CD workflows.

### DIFF
--- a/lite_plugins
+++ b/lite_plugins
@@ -1,1 +1,2 @@
-basicmessage_storage
+jwt_vc_json
+mso_mdoc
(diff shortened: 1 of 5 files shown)


lite_plugins
```diff
@@ -0,0 +1,2 @@
+jwt_vc_json
+mso_mdoc
```
jamshale, line 2: Pre-added the upcoming lite plugins

dbluhm (approved):

PR #1057, opened 2024-09-24
Add presentation for SD-JWTs

mepeltier:
This PR adds in support for presenting SD-JWTs. The demo has currently been adjusted to only generate SD-JWT presentation requests on the presentation side, so that we could get some quick testing in.
Initial tests are successful, but we do want to modify the demo so that there's an option for either type of proof request, in the same way that the demo offers options for issuing credentials.


jamshale (approved):
Can't contribute much to this review other than it looks well done 👍

dbluhm (approved):

PR #1075, opened 2024-10-02
Prepare for OWF move.

jamshale:
Updates all plugins that are upgradable to 1.1.0 and updates the package name in their pyproject.toml and their poetry.lock files. Updates all the references in the plugin source code with the new package.

Manually updates the RELEASES.md file. I think I'll have to manually create the release after as the repo token still doesn't have the create PR and releases permissions.


MAINTAINERS.md
```diff
@@ -2,40 +2,14 @@
 
 ## Maintainer Scopes, GitHub Roles and GitHub Teams
```
jamshale, line 4: I'm not exactly sure what to do with this file. Tried to copy what was done in aca-py, but still some hyperledger references.

swcurran: I can do a follow up PR to address it.  I think I know what is needed.

setupDocs.sh
```diff
@@ -16,9 +16,9 @@ cp LICENSE docs/LICENSE.md
 
 # Generate the main part of the mkdocs.yml file
 cat <<-EOF >${MKDOCS}
-site_name: Hyperledger Aries Cloud Agent Python Plugins
-repo_name: hyperledger/aries-acapy-plugins
-repo_url: https://github.com/hyperledger/aries-acapy-plugins
+site_name: A Cloud Agent Python Plugins
```
esune, line 19: What about `Python Cloud Agent Plugins`?

esune (approved):
Looks good! Left a comment about the site docs name generator script, but it could be resolved later.

swcurran:
Yay!  Awesome work, Jamie!

PR #1133, opened 2024-10-17
feat: Return the generated openid URLs for credential offers

TheTechmage:
Instead of forcing the consumer to generate the URI themselves, it would be handy to generate it for them.

cc: @dbluhm


dbluhm (approved):

dbluhm:
Checks failing; can you look into that?

TheTechmage:
They're failing because the API response changed 🤔 Which makes sense, but I also wanted a second opinion on the change. I can fix the tests tomorrow

jamshale (approved):
LGTM

Issue #1163, opened 2024-10-29
OpenID4VP SD-JWT is not verifying the holder signature (KB-JWT)


gmulhearn:
I believe the SD-JWT cred processor `verify_presentation` is not actually checking the holder's signature (KB-JWT) on the presented SD-JWT. Meaning the SD-JWT holder does not have to prove they are bound to the SD-JWT. 

I tested this locally by using a malicious client that intentionally malforms the signature bytes of the KB-JWT it produces. The acapy plugin still reported verified=true (credo reported verified=false with logs that indicate the signature was invalid).

I believe this happens because:
* verify_presentation calls sd_jwt_verify with only 2 params; `profile, presentation`: https://github.com/openwallet-foundation/acapy-plugins/blob/main/oid4vc/sd_jwt_vc/cred_processor.py#L205
* `sd_jwt_verify` has 2 additional optional params, expected_aud & expected_nonce. so verify_presentation sets those as `None`
* sd_jwt_verify initializes SDJWTVerifierACAPy with those null values for self.expected_aud & self.expected_nonce
* sd_jwt_verify then calls SDJWTVerifierACAPy.verify(), which checks the SD-JWT VC signature, and then only IF `if self.expected_aud or self.expected_nonce:` does it check the KB JWT:
```
 if self.expected_aud or self.expected_nonce:
            if not (self.expected_aud and self.expected_nonce):
                raise ValueError(
                    "Either both expected_aud and expected_nonce must be provided "
                    "or both must be None"
                )
            await self._verify_key_binding_jwt(
                self.expected_aud,
                self.expected_nonce,
            )
```
* meaning the KB-JWT signature (holder binding/proof) is not checked

dbluhm:
@mepeltier @TheTechmage fyi

dbluhm:
This should now be solved by #1172

PR #1172, opened 2024-11-06
OID4VC: Verify KB-JWT

TheTechmage:
In an attempt to fix the issue raised in #1163, we should be storing the nonce generated for presentations then checking it later on. Now, we are storing it, then passing it down to the verification logic (along with the configured endpoint/client ID) to verify the KB-JWT.


dbluhm (approved):
LGTM! Thanks!

PR #1182, opened 2024-11-15
Minor Bug Fixes

mepeltier:
In our testing, we've come across a few minor bugs that this PR intends to sort out

### DIFF
--- a/oid4vc/sd_jwt_vc/cred_processor.py
+++ b/oid4vc/sd_jwt_vc/cred_processor.py
@@ -74,7 +74,7 @@ class SdJwtCredIssueProcessor(Issuer, CredVerifier, PresVerifier):
         assert supported.format_data
         assert supported.vc_additional_data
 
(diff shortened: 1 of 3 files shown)


dbluhm (approved):

oid4vc/sd_jwt_vc/cred_processor.py
```diff
@@ -74,7 +74,7 @@ async def issue(
         assert supported.format_data
         assert supported.vc_additional_data
 
-        sd_list = supported.vc_additional_data.get("sd_list", [])
+        sd_list = supported.vc_additional_data.get("sd_list") or []
```
jamshale, line 77: Interesting. Thought this would work.

dbluhm: Indeed; subtle nuance is that `.get` will return a None if `sd_list` is present and null. We always want to treat it as a list, even if a null is given, so we have to do the `or` syntax here.

PR #1260, opened 2024-12-05
Release v1.1.1 Upgrades

jamshale:
Because of the removal of the in memory wallet all the unit tests failed so the automatic process didn't work because none were allowed to upgrade.

Updates all the unit tests to use the `create_test_profile` util.

Also another somewhat breaking change I forgot about is the `INDY_ISO8601_DATETIME_EXAMPLE` and such got changed to `ISO8601_DATETIME_EXAMPLE` because they aren't indy specific.

Also the multitenant manager module got refactored and any plugins which used it may have breaking changes.

### DIFF
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -1,4 +1,25 @@
 # Plugin Release Status
+## ACA-Py Release 1.1.1
+
+| Plugin Name | Supported ACA-Py Release |
+| --- | --- |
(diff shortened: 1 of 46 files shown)


swcurran (approved):

jamshale:
Some integration tests are failing because of the removal of the in memory wallet needing wallet name and key. Should be an easy fix.

PR #1307, opened 2024-12-20
feat: plugin for cheqd method support

sownak:
PR for new Plugin to enable support for `did:cheqd` in ACA-Py. 
- Unit Tests coverage > 90%
- Integration Tests included
- Postman collection also included

**NOTE**
Only open item is that this refers to a develop branch for the core ACAPY changes, this will be updated when the latest version of ACAPY is released.

Also includes contributions by @DaevMithran


jamshale:
So the auto create plugin script you used needs to be updated. Originally I thought most of the plugins were going to be protocols.

You can delete the `definition.py` file and then move the plugin code up a level from the `v1_0` folder. 

Then in places where the agent uses the `--plugin` config the plugin name would be `cheqd` instead of `cheqd.v1_0`.

Sorry about that. Hope it's not too much work to update, and I'll try and update the script to avoid this for the next person.

jamshale:
Did a quick over and it's looking really good 👍 Will try and find some time to go over it closer.

cheqd/pyproject.toml
```diff
@@ -0,0 +1,92 @@
...
+authors = []
+
+[tool.poetry.dependencies]
+python = "^3.12"
+cryptography = "<38.0.0"
+
+# Define ACA-Py as an optional/extra dependancy so it can be
+# explicitly installed with the plugin if desired.
+# acapy-agent = { version = "~1.1.0", optional = true }
+acapy-agent = {git = "https://github.com/cheqd/aries-cloudagent-python.git", branch = "develop"}
```
jamshale, line 14: I guess this can't be updated until the new acapy version gets released. This is fine for now. It will get updated during the next release process.

jamshale (approved):
I have no problem merging this as is. Awesome work!

swcurran:
A question came up the other day about the compatibility of this plugin with other implementations -- notably in Credo and Aries VCX.  My understanding is that the implementations are compatible correct?  It would be nice to have something about that in the documentation.  As well, @dbluhm will be adding a note about a technique he uses for basic interop testing. I'm hoping that will be easy enough to implement that we can add a couple of interop tests directly here.

swcurran:
@sownak, @DaevMithran -- would you be interested in doing a walkthrough of the Cheqd.io plugin at the next ACA-Pug -- next Tuesday 8:00 Pacific / 17:00 Central Europe?

DaevMithran:
@swcurran Sure! we can. Thanks

PR #1335, opened 2024-12-31
Acapy plugin integrating Hedera for storing DIDs/AnonCreds

paulo-caldas-code:


swcurran:
Thanks for the conbtribution! Probably won’t get looked at until next week.  @jamshale @dbluhm — heads up about this one.

paulo-caldas-code:
Apologies, by my mistake we weren't fully good to go with this pull request, but we intend to do so soon enough. I'll open a new one when that is the case. In the meantime, please disregard.

PR #1377, opened 2025-01-09
[WIP] Initial commit - Webvh

jamshale:
WIP - Adds basic did create/update/deactivate support with the `didwebvh-server-py` project.

Includes endorsement protocol, initial integration testing and some initial unit tests.


webvh/webvh/routes.py
```diff
@@ -0,0 +1,168 @@
...
+from .did.exceptions import (
+    ConfigurationError,
+    DidCreationError,
+    DidUpdateError,
+    EndorsementError,
+)
+from .did.manager import DidWebvhManager
+
+
+class WebvhOptionsSchema(OpenAPISchema):
```
PatStLouis, line 23: is this request payload following a model we see with other did registration methods or is it specific to webvh?

webvh/webvh/did/manager.py
```diff
@@ -0,0 +1,635 @@
...
+            response = await http_session.post(
+                server_url,
+                json={"didDocument": endorsed_document},
+                ssl=False,
+            )
+            response_json = await response.json()
+            if response.status == http.HTTPStatus.BAD_REQUEST:
+                raise DidCreationError(response_json.get("detail"))
+
+            log_entry = response_json.get("logEntry")
```
PatStLouis, line 319: This will change, the posting of the did document will no longer return the log entry. Instead the client must create it itself. However, I will have a helper call available by calling GET `http://example.com/my/identifier`. If this call is made after a did was registered but the log wasn't initialized, it will return a proposed initial log entry for the client to sign. However, the client should be responsible for creating this initial did log entry based on the parameters it wishes to set.

webvh/webvh/did/manager.py
```diff
@@ -0,0 +1,635 @@
...
+                self.profile,
+                Event(
+                    f"endorsement_response::{endorsed_document['id']}",
+                    {"document": resolved_did_doc["did_document"], "metadata": metadata},
+                ),
+            )
+
+        return resolved_did_doc
+
+    async def update(self, options: dict, features: dict):
```
PatStLouis, line 389: currently not implemented on the didwebvh server. The route exists, but will return a 501

webvh/webvh/did/manager.py
```diff
@@ -0,0 +1,635 @@
...
+
+            resolver = session.inject(DIDResolver)
+
+        return (
+            await resolver.resolve_with_metadata(
+                self.profile, document_state.document["id"]
+            )
+        ).serialize()
+
+    async def deactivate(self, options: dict):
```
PatStLouis, line 454: currently not implemented on the didwebvh server. The route exists, but will return a 501

webvh/webvh/routes.py
jamshale: Umm. This is super basic right now specific to `webvh`. I was mostly waiting until I knew exactly what options would be available. Currently you just provide the namespace and an optional identifier.

It tries to be the same as what cheqd did with an options and features object, as this had been discussed briefly. It's very much open for changes.

webvh/webvh/did/manager.py
jamshale: Ok. The way this was working was by creating the next custom log entry and posting it to the /namespace/identifier endpoint.

Do we need another endpoint for this?

PatStLouis: updates will be using the `PUT` method

webvh/webvh/routes.py
PatStLouis: ok good to know. I would envision having a `parameters` key in there where the user can set some boolean flags for `prerotation` and `portable` as well as the witness parameters. The options key seems fine, maybe just change `features` with `parameters` ?

We should also discuss how we plan to handle versions of webvh which may or may not become a nightmare.

jamshale:
Update/Deactivate have some old implementation still but aren't finished. Prerotation is also not implemented atm.

Integration test will fail until the server changes are merged to main.

Endorsement requires the endorser agent creating a key with a `kid` that equals the server url. Then the server needs the verify the public key. Currently the integration tests use insecure seed to automate this set up.

jamshale:
Integration test is failing because it's not using an https domain. Thinking about how to get around this with the base library.

Also, currently the ssl checks in the plugin are turned off. I'll need to update that.

webvh/webvh/did/operations_manager.py
```diff
@@ -0,0 +1,496 @@
...
+                raise DidCreationError(
+                    "Invalid response from Webvh server requesting identifier"
+                )
+
+    async def _get_or_create_author_key(self, did):
+        async with self.profile.session() as session:
+            try:
+                key_info_author = await MultikeyManager(session).create(
+                    alg="ed25519",
+                    kid=f"{did}#author",
```
PatStLouis, line 102: We will likely have to revisit this when we look at key rotation, maybe put a todo comment in here, in the webvh this would likely be called the "update Key" or "authorized Key:
https://identity.foundation/didwebvh/next/#authorized-keys

webvh/webvh/did/operations_manager.py
```diff
@@ -0,0 +1,496 @@
...
+                key_info_author = await MultikeyManager(session).from_kid(f"{did}#author")
+
+        return key_info_author
+
+    async def _create_author_signed_registration_document(
+        self, did, author_key_info, expiration, domain, challenge
+    ):
+        async with self.profile.session() as session:
+            return await DataIntegrityManager(session).add_proof(
+                DIDDocument(
```
PatStLouis, line 114: I'm ok with this, but this is making the decision to add the authorized key as a verification method. This may or may not be desirable for security reasons. I'm ok with leaving it as is and including a comment to revisit this and discuss with the acapy community.

webvh/webvh/did/operations_manager.py
```diff
@@ -0,0 +1,496 @@
...
+                author_key_info=event.payload.get("author_key_info"),
+                parameters=event.payload.get("metadata", {}).get("parameters"),
+            )
+
+    async def create(self, options: dict):
+        """Register identities."""
+
+        server_url = get_server_info(self.profile)
+
+        namespace = options.get("namespace")
```
PatStLouis, line 160: maybe have a default namespace if non is provided?
`options.get("namespace", "default")`

webvh/webvh/did/operations_manager.py
```diff
@@ -0,0 +1,496 @@
...
+        session: ProfileSession,
+        domain: str,
+        namespace: str,
+        identifier: str,
+        parameters: dict,
+        author_key_info: dict,
+    ):
+        initial_doc = {
+            "@context": "https://www.w3.org/ns/did/v1",
+            "id": r"did:webvh:{SCID}:" + domain + ":" + namespace + ":" + identifier,
```
PatStLouis, line 213: I like to use f strings:
`f"{domain}:{namespace}:{identifier}"`

webvh/webvh/did/operations_manager.py
```diff
@@ -0,0 +1,496 @@
...
+        )
+
+        # Add author proof to the log entry
+        return await DataIntegrityManager(session).add_proof(
+            doc_state.history_line(),
+            DataIntegrityProofOptions(
+                type="DataIntegrityProof",
+                cryptosuite="eddsa-jcs-2022",
+                proof_purpose="assertionMethod",
+                verification_method=f"did:key:{author_key_info.get('multikey')}#{author_key_info.get('multikey')}",
```
PatStLouis, line 232: I would rename `author` with `authorized` or `update` to use the webvh spec linguo

webvh/webvh/did/tests/test_endorsement_namager.py
```diff
@@ -0,0 +1,126 @@
...
+            "plugin_config",
+            {"did-webvh": {"role": "endorser"}},
+        )
+        await EndorsementManager(self.profile).auto_endorsement_setup()
+        assert not mock_get_active_endorser_connection.called
+
+    async def test_auto_endorsement_setup_as_author_no_server_url(self):
+        self.profile.settings.set_value(
+            "plugin_config",
+            {"did-webvh": {"role": "author"}},
```
PatStLouis, line 33: I would set the role as the "controller", as in the "controller" of the did document.
https://www.w3.org/TR/did-core/#did-controller

webvh/webvh/did/utils.py
```diff
@@ -0,0 +1,25 @@
+"""Utilities for shared functions."""
+
+from acapy_agent.core.profile import Profile
+
+from .exceptions import ConfigurationError
+
+
```
PatStLouis, line 7: maybe add the http url decoding function here?

webvh/webvh/routes.py
```diff
@@ -0,0 +1,205 @@
...
+
+
+class WebvhDeactivateSchema(OpenAPISchema):
+    """Request model for deactivating a Webvh DID."""
+
+    id = fields.Str(
+        required=True,
+        metadata={
+            "description": "ID of the DID to deactivate",
+            "example": "did:webvh:prod:1",
```
PatStLouis, line 77: this example doesn't seem right

PatStLouis (commented):
First round of review/comments

jamshale:
The integration test for auto endorsing isn't working even though it's working manually. The test still allows the initial log to get posted so I removed the assertion for now.

webvh/webvh/did/utils.py
```diff
@@ -0,0 +1,63 @@
...
+    Args:
+        json_obj (dict): The JSON object to encode.
+        multibase_encoding (str): The multibase encoding format (e.g., 'base58btc').
+
+    Returns:
+        str: The multibase-encoded hash of the JSON object.
+    """
+    # Serialize JSON object to a compact string
+    json_str = json.dumps(json_obj, separators=(",", ":"))
```
PatStLouis, line 55: The json should be canonicalized with a conformant library. Here's one:
https://github.com/titusz/jcs

webvh/webvh/did/utils.py
```diff
@@ -0,0 +1,63 @@
...
+
+    # Compute SHA-256 hash of the JSON string
+    json_hash = hashlib.sha256(json_str.encode("utf-8")).digest()
```
PatStLouis, line 58: Might want to use multi hash here:
https://www.w3.org/TR/controller-document/#multihash

PR #1447, opened 2025-01-24
Feature/connection protocol

jamshale:
Working off of Daniels branch. Updates the plugin to work with the latest acapy.

Maybe want to add some more of the scenario tests that were removed to the integration tests here. Currently just has one simple integration test to create a basic connection.

See original PR https://github.com/openwallet-foundation/acapy-plugins/pull/925


dbluhm (approved):

swcurran:
w00t!!!! Great work.  @ianco — once this is merged, would love to see the if the connections tests on OATH that started failing because I merged the removal PR in ACA-Py are fixed by loading in this plugin.  Even getting feedback on a local run would be helpful.  I think in OATH, we could just always load the plugin for all ACA-Py runs, or (better) load it only when doing a connections test.

jamshale:
Weird. Daniel doesn't have approval rights now. He should have....

swcurran:
AFACS he has approval rights.  The PR was ready to merge when I came to the page.

jamshale:
👍OK. Not sure why it was blocked when I looked at it. Thanks.

jamshale:
@swcurran I will look into the OATH stuff with the plugin. I think @ianco has enough and is on vacation in a week.

ianco:
> w00t!!!! Great work. @ianco — once this is merged, would love to see the if the connections tests on OATH that started failing because I merged the removal PR in ACA-Py are fixed by loading in this plugin. Even getting feedback on a local run would be helpful. I think in OATH, we could just always load the plugin for all ACA-Py runs, or (better) load it only when doing a connections test.

The easiest approach is to install the plug-in as part of the aca-py backchannel build and then always load it when the agent is started.

We have some infrastructure to restart the agent with custom parameters for specific tests but I've had some issues on occasion when doing this - the most "stable" approach is just to add the plug-in to the default config.

PR #1526, opened 2025-02-13
Add plugin to support Hedera DID Method and AnonCreds registry

AlexanderShenshin:
This PR contains Hedera plugin, enabling support for [Hedera DID Method](https://github.com/hashgraph/did-method/blob/master/hedera-did-method-specification.md) and AnonCreds registry.

Includes automated tests (unit + integration, coverage > 95%) and interactive demo for AnonCreds flow (in a form of console apps).

The implementation is based on [Hiero DID SDK Python](https://github.com/hiero-ledger/hiero-did-sdk-python).
Please note that it's currently added as a Git dependency, will be changed to PyPi package once it's released.


jamshale (commented):
This was very well done. I couldn't find anything done incorrectly and it was very cleanly written.

The integration test are failing in github actions but it's possible there's an issue with the project overall and not this plugin. I'll have a look. When that's figured out I'll approve and merge.

jamshale:
So we run the integration tests automatically with the github actions CI/CD pipeline. This is because we try and maintain dependency upgrades and compatibility with acapy. It doesn't seem like the integration tests can run with an operator_id and operator_key?

I'm not sure if there's a way to run these tests publicly? If not we'll need a basic docker compose that just runs and completes and we can have these tests in a different named docker-compose file like done in https://github.com/openwallet-foundation/acapy-plugins/tree/main/oid4vc/integration.

Or we could add something to the project to skip the integration tests for this plugin and rely on the unit tests and manually running the integration tests occasionally.

AlexanderShenshin:
Hi @jamshale , thanks for feedback!

Good point, these integration tests run against Hedera Testnet, which requires operator credentials.
I think generally it would be good to have ability to run such tests in this repo. In case of Hedera plugin, it's possible to provide operator config via env variables.

We could create GH Action secrets and update GH workflow a bit to pass these to environment. For example, in integration tests run job we can add:
```
env:
  HEDERA_OPERATOR_ID: ${{ secrets.HEDERA_OPERATOR_ID }}
  HEDERA_OPERATOR_KEY: ${{ secrets.HEDERA_OPERATOR_KEY }}
```

However, we have generic CI/CD workflow here, so I'm wondering if it will look good for a general approach. What do you think?

AlexanderShenshin:
Meanwhile, I added testnet operator credentials for integration tests specifically.
These are not sensitive, I suggest that we remove them once we have env config setup or decide on other option.

jamshale:
@AlexanderShenshin Thank you. I will merge it like this for now. We could change it to use secrets but I don't have full control over the repo to manage this myself. If you'd still like to go that way let me know and I can organize it with someone who can.

PR #1530, opened 2025-02-18
oid4vc: cred_offer_uri

Athan13:
On creation, credential offers can now be returned to the controller by reference as well as by value, using the "credential_offer_uri" field (see Section 4.1.3 of the [spec](https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0-ID1.html#name-sending-credential-offer-by-)). Works for both JWT and SD-JWT credentials. [Duplicated from the Indicio fork].


jamshale (approved):
LGTM 👍

PR #1531, opened 2025-02-20
fix(hedera): Update Hiero SDK dependency and improve README

AlexanderShenshin:
- Updated Hiero SDK dependency (previously used "dev" package was removed)
- Removed relative path references from Hedera plugin README, these are not working properly when published to GH pages


swcurran (approved):

AlexanderShenshin:
@swcurran Re-requesting the review - converted PR to draft after noticing some issues with tests (required dependency update).

But now it's fixed and good to go.

swcurran (approved):

PR #1542, opened 2025-02-24
Webvh did method

jamshale:
This is this initial webvh plugin. 

Features:

- Basic did creation
- Anoncreds with attested resources support
- witnessing for server authentication via didcomm

Todo(s):

- did update with prerotation and portablity (wip - initial did update endpoint is available but not fully operational)
- same thing with deactivation
- Additional anoncreds integration testing. Currently a single basic test.
- Interop testing when credo support added


jamshale:
I'm going to merge this. I think it's good enough to get in as an initial version.

PR #1544, opened 2025-02-28
feat: Use acceptheaders in cheqdResolver for dereferencing

DaevMithran:


cheqd/cheqd/anoncreds/registry.py
```diff
@@ -179,7 +183,6 @@ async def register_schema(
             LOGGER.debug("Existing schema %s", existing_schema)
             # update if schema exists
             if existing_schema is not None:
-                LOGGER.debug("UPDATING SCHEMA")
                 cheqd_schema = ResourceUpdateRequestOptions(
                     options=Options(
                         name=resource_name,
```
sownak, line 188: @DaevMithran  Does "updateResource" work when the same name and type is passed?

sownak:
@DaevMithran Please also update the images to ghcr.io/cheqd/did-registrar:production-latest in the docker-compose files

cheqd/cheqd/anoncreds/registry.py
DaevMithran: The revocation tests already handle it. Added duplicate test for schemas too

sownak (approved):

jamshale (approved):

Issue #1543, opened 2025-02-24
BUG: 500 error if presentation exchange (VS DCQL) is used for SD-JWT


timbl-ont:
When using PEX it looks like dcql_query is undefined - error on this line:

https://github.com/openwallet-foundation/acapy-plugins/blob/def15370a054321385866b1812651dda92ad719f/oid4vc/oid4vc/public_routes.py#L525

` File "/usr/src/app/oid4vc/public_routes.py", line 503, in get_request
demo-issuer-1            |     if dcql_query:
demo-issuer-1            |        ^^^^^^^^^^
demo-issuer-1            | UnboundLocalError: cannot access local variable 'dcql_query' where it is not associated with a value`

FYI: @mepeltier

Issue #1446, opened 2025-01-22
OID4VC: Demo creates new credential_supported metadata entry for each issuance (BUG)


timbl-ont:
Each time the demo is used to issue a new credential, a call is made to create a new credential_supported entry. The effect is that the metadata will include the same credential type multiple times (after the first issuance).

For some wallets this appears to work ok, but it breaks the Sphereon wallet, which attempts to make a credential request for each entry of the issued credential type in the credential metadata. Effectively it only works the first time a credential is issued.

To fix this, the demo will need to persist a flag to indicate that the metadata has been created and also persist the supported_cred_id for subsequent issuance. 

Note: the holder binding needs to be set to did:jwk to support the Sphereon wallet. Otherwise did:key is used, which is not supported by the plugin.

Issue #1443, opened 2025-01-22
OID4VC: Ability to patch credential metadata


timbl-ont:
The current API can create and delete credential metadata via the credential-supported endpoint. Each supported credential type is associated with a supported_cred_id which is used by the controller to issue a credential of that type.

Any updates of the metadata require a new entry (and old entry deleted) which changes the supported_cred_id. This then requires that the controller be updated and redeployed.

The addition of PATCH to the credential-supported endpoint would enable the updating of the metadata without changing the supported_cred_id. For simplicity it could be a wholesale replacement of the JSON configuration.

Issue #1429, opened 2025-01-16
OID4VC plugin needs some readme updates


dbluhm:
There's text in the plugin README that was true at the time of our first contribution but is no longer true after subsequent updates to the plugin. We should bring it up to date.

PR #1577, opened 2025-03-03
OID4VC: Bugfixes and API Updates

mepeltier:
This PR includes a number of somewhat miscellaneous updates and fixes for the OID4VC plugin.
Some highlights:
- Fix the bug reported in #1543 
- Fix the Credential Supported recreation reported in #1446 
- Add ability to patch `PresDef` and `SupportedCredential` records (#1443)
  - As a note, these new `PUT` endpoints perform a complete overwrite of the existing record, meaning that any optional attributes that aren't supplied in the new request will be `None`, rather than maintaining whatever value they were previously
- Some README updates are included (#1429)
  - @dbluhm were there other specific updates you were looking to see? 
- Better checking on the "generic" `/supported-credential/create` endpoint
  - some values are expected to only be present for specific formats, and shouldn't both be present at the same time, so we now perform some collision checking there. It didn't *break* anything to allow both attributes, but it could get confusing for debugging/analysis purposes

### DIFF
--- a/oid4vc/README.md
+++ b/oid4vc/README.md
@@ -193,7 +193,7 @@ The Plugin expects the following configuration options. These options can either
 
 ### Creating Supported Credential Records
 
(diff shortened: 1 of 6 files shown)


jamshale (approved):
LGTM 👍

PR #1607, opened 2025-03-25
:construction_worker: Dump logs if tests fail

ff137:
To help with debugging failing integration runs


jamshale (approved):
Thanks. Been meaning to do this.

PR #1609, opened 2025-03-26
:arrow_up: Bump docker-compose-wait to latest

ff137:
From 2.7.2 to 2.12.1
https://github.com/ufoscout/docker-compose-wait/releases

### DIFF
--- a/.github/workflows/pr-integration-tests.yaml
+++ b/.github/workflows/pr-integration-tests.yaml
@@ -92,7 +92,28 @@ jobs:
       - name: Run Integration Tests
         id: integration-tests
         run: |
+          # Process redis_events first if it exists in the list
+          if [[ "${{ steps.changed-plugins.outputs.changed-plugins }}" == *"redis_events"* ]]; then
+            echo "Running integration tests for redis_events first"
+            cd redis_events/integration
+            docker compose build
+            if ! docker compose up --exit-code-from tests; then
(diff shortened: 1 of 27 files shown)


ff137:
@jamshale I think I got to the bottom of it, with this change: https://github.com/openwallet-foundation/acapy-plugins/pull/1609/commits/66003fdd57aacc8976e767081f294d0c5921aa5b

The logs showed:
```
2025-03-26T20:52:43.0380408Z cluster         | [OK] All 16384 slots covered.
2025-03-26T20:52:43.1588428Z 
2025-03-26T20:52:43.1590900Z [Kcluster exited with code 0
2025-03-26T20:52:43.1594626Z Aborting on container exit...
2025-03-26T20:52:43.1652401Z  Container juggernaut  Stopping
```

And with GPT as sidekick, that was pointed out as the problem - that the redis-cli cluster command succeeds and exits, causing the other containers to exit.

This is related to the change introduced here: https://github.com/openwallet-foundation/acapy-plugins/pull/1607. 
Running with `docker compose up --exit-code-from tests` instead of the previous `docker compose run tests` is now different in how it handles exit codes. `docker compose run tests` doesn't care about other containers. `up tests` instead prompts the whole stack to run and it monitors all containers.

May be different tweaks to avoid this scenario, but it appears that keeping redis cluster command "alive", with the tail command at the end, does the trick

jamshale (approved):
Great! thanks.

PR #1654, opened 2025-04-03
[WebVH] did update with pre-rotation

PatStLouis:
Improved some of the did registration / creation and added did update feature.

Will look at the tests


webvh/webvh/did/utils.py
```diff
@@ -1,42 +1,42 @@
...
-def key_to_did_key_vm(multikey: str):
-    """Transform a multikey to a did key verification method."""
-    return f"did:key:{multikey}#{multikey}"
+def create_alias(identifier: str, purpose: str):
+    """Get static alias."""
+    return f"webvh:{identifier}{ALIAS_PURPOSES[purpose]}"
 
 
 def server_url_to_domain(server_url: str):
     """Replace %3A with : if domain is URL encoded."""
```
esune, line 36: Any reason to just handle this url-encoded character, rather than the whole set (i.e.: returning the full url-decoded url)?

webvh/webvh/did/witness_manager.py
```diff
@@ -235,3 +259,31 @@ async def attest_did_request_doc(self, entry_id: str) -> dict[str, str]:
...
+
+    async def create_invitation(self, alias=None, label=None) -> str:
+        """Create a witness invitation."""
+        witness_key = await self.get_witness_key()
+        if not witness_key:
+            raise WitnessError("No active witness key.")
+
+        try:
+            invi_rec = await OutOfBandManager(self.profile).create_invitation(
+                hs_protos=[HSProto.get("https://didcomm.org/didexchange/1.0")],
```
esune, line 281: Do we ever plan on supporting didexchange 1.1?

esune (commented):
Good progress! Will do another pass once officially ready for review.

webvh/webvh/did/witness_manager.py
swcurran: I would hope that we do.  At minimum we need to accept incoming both 1.0 and 1.1, and I think (but not certain) that we should be sending 1.1.  @TelegramSam, @TheTechmage  — do you recall where we are with DID Exchange.  What does Credo-TS support?

PatStLouis: I was just using what was in the examples, I can update the array to have both values it shouldn't be an issue.

webvh/webvh/did/utils.py
PatStLouis: I think the goal is only to cater to the port delimiter being encoded rather than all characters

esune:
General note/comment: as I deployed the plugin in a test environment, it seems like the startup argument to activate it is not following the pattern we have for the other plugins: we usually use `--plugin name.v1_0`, but that did not work for the webvh plugin and I had to use `--plugin webvh` instead. Maybe related to the fact it was not actually released?

jamshale:
> General note/comment: as I deployed the plugin in a test environment, it seems like the startup argument to activate it is not following the pattern we have for the other plugins: we usually use `--plugin name.v1_0`, but that did not work for the webvh plugin and I had to use `--plugin webvh` instead.
> 
> The `webvh` plugin does not follow the same convention specifying the plugin version as other plugins - we may want to address that (maybe as a separate PR where files are "moved only" rather than with in-progress work.

The `.v1_0` naming pattern was from the original plugins that came from traction. I kept it a long time ago and think it can still make sense for protocol focused plugins but it's been largely abandoned.

esune:
> > General note/comment: as I deployed the plugin in a test environment, it seems like the startup argument to activate it is not following the pattern we have for the other plugins: we usually use `--plugin name.v1_0`, but that did not work for the webvh plugin and I had to use `--plugin webvh` instead.
> > The `webvh` plugin does not follow the same convention specifying the plugin version as other plugins - we may want to address that (maybe as a separate PR where files are "moved only" rather than with in-progress work.
> 
> The `.v1_0` naming pattern was from the original plugins that came from traction. I kept it a long time ago and think it can still make sense for protocol focused plugins but it's been largely abandoned.

Ok, it sounds like we may be good for this specific plugin then.?

jamshale:
This looks good other than the failing tests which might be related to saving a config in the tests with the old table name.

jamshale (commented):
Ya. I think this works fine :+1: I'll rebase the integration test fixes later today.

jamshale (approved):

PR #1671, opened 2025-04-10
:art: Rename Anoncreds to AnonCreds

ff137:
Standardises the naming convention, in alignment with https://github.com/openwallet-foundation/acapy/pull/3573

___
Keeping in draft until above PR is merged and in next release

### DIFF
--- a/hedera/hedera/anoncreds/registry.py
+++ b/hedera/hedera/anoncreds/registry.py
@@ -19,7 +19,7 @@ from acapy_agent.anoncreds.base import (
     SchemaResult,
 )
 from acapy_agent.anoncreds.events import RevListFinishedEvent
-from acapy_agent.anoncreds.models.schema_info import AnoncredsSchemaInfo
+from acapy_agent.anoncreds.models.schema_info import AnonCredsSchemaInfo
 from acapy_agent.core.event_bus import EventBus
 from acapy_agent.wallet.base import BaseWallet
 from hiero_sdk_python import Client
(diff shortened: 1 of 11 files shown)


swcurran:
Do these change only calls into ACA-Py, or API calls into the plugins?  If the latter, have to let the users of those plugins aware of this.

Either way -- heads up about this @DaevMithran @sownak (Cheqd Plugin), @AlexanderShenshin (Hedera Plugin) and @jamshale @PatStLouis (did:webvh plugin).

swcurran:
FYI--the related ACA-Py issue has been merged.

ff137:
I imagine it's only imports that have to be renamed. And then if anybody is using an OpenAPI-generated client (like we do), then models will be renamed when they regenerate the client. Otherwise it's an internal change and everything should work the same

hedera/hedera/anoncreds/registry.py
```diff
@@ -19,7 +19,7 @@
     SchemaResult,
 )
 from acapy_agent.anoncreds.events import RevListFinishedEvent
-from acapy_agent.anoncreds.models.schema_info import AnoncredsSchemaInfo
+from acapy_agent.anoncreds.models.schema_info import AnonCredsSchemaInfo
```
jamshale, line 22: I think this one might have been missed? 

`cheqd/anoncreds/tests - ImportError: cannot import name 'AnonCredsSchemaInfo' from 'acapy_agent.anoncreds.models.schema_info' (/home/runner/.cache/pypoetry/virtualenvs/cheqd-amkBMEd--py3.12/lib/python3.12/site-packages/acapy_agent/anoncreds/models/schema_info.py)`

ff137: It's all renamed on acapy main branch, so should be in 1.3.0 or next rc release

jamshale:
Any updates on this? Looks like it's almost there...

hedera/hedera/anoncreds/registry.py
jamshale: Oh. I see. This is waiting for the release and then we'll merge it 👍

jamshale:
This is included in main now.

PR #1687, opened 2025-04-15
Feature: Add route to manually bind a status list entry

TheTechmage:
When I made this change, there was no way to bind/assign a status list entry to a credential. Without modifying ACA-Py or acapy-plugins with the changes needed to assign the status list, I found it impossible to use the feature. While I understand that the intent is to have it happen automatically, there are cases where automatically assigning status lists is not currently possible without making changes to ACA-Py. For example: When using status lists with Anon-creds or JSON-LD issuance which are part of ACA-Py core.

This change adds an API route that calls `assign_status_entries` to handle the mentioned use-cases.


TheTechmage:
> Thank you for submitting the PR. The **assign_status_entries** function is designed for Python function calls within OID4VCI credential processors to reduce controller overhead and avoid round-trip REST API calls between ACA-Py plugins. The new endpoint wraps this function, generating a W3C or IETF status JSON segment for credentials.

That seems like an accurate assessment from my understanding. In our specific use-case, we wanted to use status lists with JSON-LD credentials which are part of acapy core, making it a tad bit more difficult to call the API in the plugin without potentially breaking acapy or making it dependent on having the plugin installed.

> For non-OID4VCI use cases, we can have an endpoint that returns a list index, entry number, and entry value? This would enable more flexible data handling by the controller, rather than providing a fixed W3C/IETF JSON segment. Thoughts?

I do believe such an [API call already exists, `/status-list/defs/{def_id}/entries`](https://github.com/openwallet-foundation/acapy-plugins/blob/0d1797dcbc00a0f4599e51289f4bd6f5a0e32b8e/status_list/status_list/v1_0/status_handler.py#L194-L206). However, while this does hand out/assign a status list entry, it doesn't create a `StatusListCred` record bound to an entry and "credential". The `StatusListCred` record is *only* created by the `assign_status_entries` method and the [`update_status_list_cred` method](https://github.com/openwallet-foundation/acapy-plugins/blob/0d1797dcbc00a0f4599e51289f4bd6f5a0e32b8e/status_list/status_list/v1_0/routes.py#L33-L36) (the method to change the status of the entry) [requires the `credential_id` to be passed in](https://github.com/openwallet-foundation/acapy-plugins/blob/0d1797dcbc00a0f4599e51289f4bd6f5a0e32b8e/status_list/status_list/v1_0/status_handler.py#L313-L321) since it's in the route URL and in the tag filter. If we don't find a `StatusListCred` record when updating the entry, [the code depending on the record's existence](https://github.com/openwallet-foundation/acapy-plugins/blob/0d1797dcbc00a0f4599e51289f4bd6f5a0e32b8e/status_list/status_list/v1_0/status_handler.py#L322-L324) will throw an exception and the entry won't be updated.

I hope all of that makes sense 😅 I'm not exactly certain that what I've done here is the correct approach. I was just trying to solve the problem in front of me (assign a status entry and update it) without breaking a lot of code. Since I'm not all to familiar with the code or it's direction I didn't want to make drastic changes that would get thrown out anyways.

Thank you for taking a look at this PR @weiiv!

weiiv:
Thanks @TheTechmage for your detailed explanation, I'm happy to accept the PR.

jamshale (approved):

PR #1721, opened 2025-05-06
:art: Update UUID validation (use stricter, canonical UUID format)

ff137:
Improves on the expected UUID format

Wanted to just drop the commas and simplify {36,36} -> {36}, but figured may as well enforce canonical UUID format, and allow upper case hex as well. Happy to remove upper case if necessary


DaevMithran:
Looks Good @ff137 !

jamshale (approved):

PR #1755, opened 2025-06-05
[WIP] Self attest with non-auto attestation

jamshale:
Moved some server request logic to a client service.

Queue self attested log entries when agent is a witness but auto-attest is false.

Thought about trying to avoid the witness using controller module but decided against it as the witness does pretty much become a controller when sending it's initial log entry to the server. Avoided circular imports by importing them only when this happens.

TODO: More automated tests


PatStLouis (dismissed):
Looks good to me, I like having a separate server client class.

jamshale:
Updated comments and changed the server client to be class instance scoped.

PatStLouis (approved):

PR #1801, opened 2025-06-11
Add watchers notification on DID updates

PatStLouis:
This PR adds a new webvh configuration flag `notify_watchers` as well as a `watchers` parameter on the DID create route.

When the controller updates a DID and the `notify_watchers` flag is set to `true`, it will notify every watcher from the DID's parameters after uploading the latest DID state to the server.


jamshale (dismissed):
Just have a question about the notify process but everything looks good.

jamshale (approved):

PR #1934, opened 2025-07-10
Status list enhancement

weiiv:
- use separate configure vars for local file path and public url
- store issuer did and status list type in status list definition records


weiiv:
addressing merge conflicts

PR #1935, opened 2025-07-10
Status list enhancement

weiiv:
- separate configurations for local file path and public URL
- store issuer did/verification-method and list type in status list definition record


jamshale (approved):
Looks good 👍

PR #1974, opened 2025-07-24
:sparkles: Cheqd: Support long resource names

ff137:
Currently the DID Registrar will fail if a resource name is longer than 64 characters. This will now automatically hash resource names longer than that limit.

Signed-off-by: ff137 <ff137@proton.me>


DaevMithran:
> @DaevMithran @sownak 
> 
> As far as we can tell, the actual resource name limit is 64 characters. Is that true?

@ff137 yes it is limited. We are hashing it in credo-ts similarly. Thanks for the PR

jamshale (approved):

jamshale (approved):

PR #2005, opened 2025-08-05
Week 32, 2025 - Library Upgrades

jamshale:
General upgrades. pytest-asyncio, ruff, aiohttp, google-python-api-client, redis


swcurran (approved):
LGTM

Random question -- should ruff be added to all of plugins at some point?

PR #2010, opened 2025-08-07
🎨 Reduce noisy logs

ff137:
Sets the Cheqd did-registrar to LOG_LEVEL: warn, to avoid spamming integration test output

___
PS: A tails-server update was just pushed, which requires --log-level args to be upper case. I notified them that lower case should be supported too. But in the meantime, I'm just changing the tails-server log level to upper case to account for that (as opposed to pinning to previous version)

### DIFF
--- a/cheqd/integration/docker-compose.yml
+++ b/cheqd/integration/docker-compose.yml
@@ -134,4 +134,4 @@ services:
       --host 0.0.0.0
       --port 6543
       --storage-path /tmp/tails-files
-      --log-level info
+      --log-level INFO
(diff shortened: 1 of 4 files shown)


cheqd/integration/docker-compose.yml
```diff
@@ -90,6 +90,9 @@ services:
       - "9080:9080"
     depends_on:
       - driver-did-cheqd
+    environment:
+      LOGGING_LEVEL_ROOT: "WARN"
+      LOGGING_LEVEL_uniregistrar: "WARN"
```
ff137, line 95: @sownak just want to share, this successfully modifies the uniregistrar log level. Copied it from how we have it in acapy-cloud :-)

sownak: I also found this :)

jamshale (approved):

PR #2011, opened 2025-08-10
Add WebVH Witness protocol webhooks

PatStLouis:
Adding webhooks to use with the endorser service. Covers both the log entry and attested resource.


jamshale (approved):
Let's get this in before I do a plugin release tomorrow.

PR #2024, opened 2025-08-13
fix: OID4VC `sd_jwt` extra renamed to `sd_jwt_vc` to resolve build failure

pradeepp88:
This update corrects the extra name in pyproject.toml in OID4VC plugin from `sd_jwt` to `sd_jwt_vc` to align with the actual package name.
The incorrect extra name caused build failures during dependency installation, and this change ensures the build process completes successfully.


jamshale:
I think it's just the lock file. You should just be able to run poetry lock in the root of the plugin and then commit the poetry.lock file.

PR #2055, opened 2025-08-26
Week 35, 2025 - Library Upgrades

jamshale:
requests, ruff, google-api-python-client, protobuf , did-webvh


jamshale:
@AlexanderShenshin Anything we can do to get the integration test hedera ledger working again. We plan on doing a plugins release tomorrow.

AlexanderShenshin:
@jamshale I've identified the problem and created an issue + PR with a fix for Hedera SDK Python.
https://github.com/hiero-ledger/hiero-sdk-python/issues/316

In short, the problem is related to deprecation of certain endpoints in Hedera network.

As of plugins release, I think the optimal way would be to ignore this if it's not a complete blocker. Once there is a new SDK version with a fix, I'll work on updating the plugin ASAP.
And if we can't ignore this for the release, I can publish a quick (forked) hotfix version tomorrow. But my optimistic expectation is that the new version of SDK will be released quite fast, maybe tomorrow as well (taking into the account severity).

swcurran (approved):

PR #2264, opened 2025-10-15
Use ACAPY's P256 key type in OID4VC Plugin

gmulhearn:
This was giving me problems in base acapy, as the oid4vc plugin was overriding the registered P256 keytype with it's own class.

Specifically i observed this problem when trying to create a did:key with P256 - acapy would report that did key does not support p256


jamshale (approved):
Makes sense to me. Nice fix.

jamshale:
@weiiv Do you have any concerns about this? I know you're heavily involved with this plugin.

weiiv:
@jamshale that looks good to me. I think OID4VC defined its own class before it was available in ACA-Py.

PR #2266, opened 2025-10-15
Publish acapy-plugins-bundle image

PatStLouis:
- Adds a Dockerfile for the bundle
- Adds a pyproject file for the bundle
- Adds a publish workflow for the bundle

On new release, ensure to update the pyproject.toml bundle version, matching the release tag.

Job will be triggered on published release, and will cancel if an existing package of that name/version exists

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -1,5 +1,5 @@
 name: Publish ACA-Py with Plugins
-run-name: Publish ACA-Py ${{ github.event.release.tag_name }} with all plugins
+run-name: Publish ACA-Py ${{ github.event.release.tag_name || github.event.inputs.version }} with all plugins
 
 on:
   release:
@@ -17,7 +17,6 @@ permissions:
 
 jobs:
   build-and-publish:
-    runs-on: ubuntu-latest
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -0,0 +1,42 @@
...
+
+# Install ACA-Py
+RUN pip install acapy-agent==${ACA_PY_VERSION}
+
+# Create acapy user
+RUN useradd -m -s /bin/bash acapy
+
+# Install plugins with --no-deps to avoid dependency conflicts
+RUN pip install --no-deps \
+    git+https://github.com/${REPO_OWNER}/acapy-plugins@${ACA_PY_VERSION}#subdirectory=basicmessage_storage \
```
jamshale: I think this works fine. I was trying to think if copying the source code and then installing would be better. I think I like doing it this way though. Much simpler.

.github/workflows/publish.yml
```diff
@@ -0,0 +1,168 @@
...
+        required: true
+        type: string
+
+permissions:
+  contents: read
+  packages: write
+
+jobs:
+  build-and-publish:
+    strategy:
```
jamshale: For most of the other workflows in the openwallet-foundation repos we have a statement `if: github.repository == 'openwallet-foundation/acapy-plugins'` so these don't get run automatically when you have a fork.

jamshale: After discussion. This is fine to leave without a conditional.

PR #2305, opened 2025-10-27
Week 44, 2025 - Library Upgrades

jamshale:
Only `ruff` and `fastapi-slim` this week.


jamshale:
cheqd tests are failing. They can be flaky sometimes. I'll try later and if it continues I'll revert changes for cheqd plugin.

PR #2364, opened 2025-11-13
Redis cache

jamshale:
Getting this updated with newest acapy version, etc....


jamshale:
I think this should be good. I needed to change some stuff with github actions and the networking of the redis subnet.

One thing I noticed that I hadn't looked into was that in clustered mode the cache didn't have the did:peer resolver cache information but did have the connection information. Not sure what's going on with that. I think we could still merge this and open a ticket to look into that in the future.

PR #2418, opened 2025-11-18
Hedera acapy-agent upgrade

jamshale:
Trying to upgrade hedera plugin and isolate the failing integration test.


jamshale:
This failure was intermittent and is passing now.

PR #2438, opened 2025-11-25
Webvh witness invitation update

PatStLouis:
This PR includes:
- A dedicated keychain class for the controller, which manages update, next, signing and witness keys.
- Added an event manager and external state management for the witness protocols
- Updated logic to fetch a witness invitation from the webvh server
- Updated startup event auto config
- Added subwallet creation event listener


swcurran:
@PatStLouis -- can this PR be closed as "won't do", or is it still needed?  Lots have happened with did:webvh since this was opened :-)

PR #2625, opened 2026-01-13
Week 3, 2026 - Library Upgrades

jamshale:
General upgrades, aiohttp, urlib3, ruff, etc..


swcurran (dismissed):
Do we need to worry about the Integration Test failure?  Looks intermittent -- running out of space so assume it's fine.  Could it be the logging level?

PR #2762, opened 2026-02-24
webvh: fix controller AnonCreds witness requests and rev reg activation

PatStLouis:
- Rely on ACA-Py event chain for revocation registry activation after witness approval
- Improve request storage and record handling for controller
- Event bus patch and attested resource protocol updates
- Integration test updates

### DIFF
--- a/webvh/integration/tests/test_anoncreds.py
+++ b/webvh/integration/tests/test_anoncreds.py
@@ -182,9 +182,7 @@ def _cred_def_id_encoded(cred_def_id: str) -> str:
 
 async def _get_active_rev_reg_id(agent, cred_def_id: str) -> str:
     """GET active revocation registry for cred def; return rev_reg_def id."""
-    path = (
-        f"/anoncreds/revocation/active-registry/{_cred_def_id_encoded(cred_def_id)}"
-    )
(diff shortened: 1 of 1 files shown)


webvh/webvh/anoncreds/registry.py
```diff
@@ -560,6 +629,375 @@ async def _get_default_verification_method(self, profile, did):
...
+                        entry.value,
+                        tags,
+                    )
+                    LOGGER.debug(
+                        "Updated cred def state to finished locally: %s", cred_def_id
+                    )
+                    # Emit CredDefFinishedEvent so revocation setup runs
+                    if tags.get("support_revocation", "False") == "True":
+                        try:
+                            schema_id = tags.get("schema_id", "")
```
jamshale: Is defaulting to an empty string the right thing to do here? Has a bit of a code smell.

esune (approved):
Only small comment - which does not need to hold up from merging - is that there are a couple source files that are REALLY HUGE (1k lines). While this is not an issue per se, I wonder what are the thoughts round some minor refactoring to reduce large source files. This may not be possible/simple since all the code appears to belong to the same context, but thought it would be worth mentioning.

PR #2967, opened 2026-04-06
Week 15, 2026 - Library Upgrades

jamshale:
ruff, aiohttp, requests

Also removed the javascript dependabot notifications for oid4vc plugin. It can be managed separately.


swcurran:
Still don't quite get it, but let's take it to the Maintainers call tomorrow.

PR #3054, opened 2026-04-22
fix: update poetry.lock from oid4vc/auth_server

thiagoromanos:
The `pyproject.toml` present in `oidc/auth-server` was changed in the last commit, but `poetry.lock` was not. I found the error running `docker compose build ` in the demo. Just running poetry lock fixed it.


weiiv:
Great thanks @thiagoromanos. @jamshale, can you review and approve? Thanks.

jamshale (approved):

PR #3146, opened 2026-05-25
Weekly Dependabot Updates 20260525

swcurran:
Attempting the weekly update.  The instructions are not yet clear to me. Here is what I did and questions:

1. Created a branch.
2. Ran the repo_manager option 6.
3. Submitted the PR

Is there anything else that should be done? Are there more complex versions of the update process? For example, in this case, all of the dependabot PRs only update the poetry.lock files -- but what if the pyproject.toml files need to be updated?

Also, I notice that #3132 updates 10 lock files across the repo.  Could we have used that?

I haven't gotten to the closing yet. Want to see if this worked before I do that.


jamshale:
Yes. We could have used https://github.com/openwallet-foundation/acapy-plugins/pull/3132 this time. I don't know what's going on with the dependabot grouping. Usually it doesn't work correctly and then sometimes it does.

In general you shouldn't really need to change pyproject.toml files unless you want to change the minor/major upgrade range.
 
I usually just use the `python repo_manager 7 idna` for example. This will target the individual library instead of a general update of all libraries.

Then you can use `python repo_manager.py 8 <lower> <upper>` to close a range of PR's to clean up the list of PR's.

swcurran:
@jamshale -- you didn't approve this PR. Does that mean I should close it and try again targeting the idna library?

swcurran:
Also, doesn't it make sense to always update all of the lock files, so all of the dependencies get updated?

Also, how does something like ruff get updated -- references to things not covered by the lock file updates.  For example, from this [weekly upgrade](https://github.com/openwallet-foundation/acapy-plugins/pull/3068/changes) when ruff and pytest got updated across the board.

jamshale:
This PR didn't actually upgrade the libraries if you look at the lock files. I think option`6` might be broken after the major poetry version that happened a while ago. It used to upgrade all libraries that could be upgraded which is more than needed and usually broke automated testing.

Option 7 works because it goes through every project and targets the specific library you define. I use this for upgrading `ruff` which has regular updates. The PR you linked was one where I manually did the major upgrade of pytest.

I think we can either close this PR and approve a merge https://github.com/openwallet-foundation/acapy-plugins/pull/3132 this week or it can have option `7` targeting `idna` applied to this PR and then we can merge it.

swcurran:
Wow...very weird.  OK. I'll try again.

PR #3181, opened 2026-06-08
Weekly Dependabot Update -- 20260608

swcurran:
@jamshale -- please review carefully -- especially the change to the `repo_manager.py` script.

Running option 7 on my machine was taking 35-60 minutes per library, so I asked Claude for how to improve that. Steps to here were:

- Run all in parallel -- except I noted that there were a lot of lock files (69 it turns out...).
- Run 6 in parallel and each group in sequence
- On first run, got several errors because of conflicting venv updates
- Suggestion was to only update the lock file and to leave the venv alone.

Updates are now on the 2-5 minutes range.  But are they correct?


swcurran:
Updates for this week:

- ruff
- uvicorn
- requests
- pytest-asyncio
- aiohttp

jamshale:
Wow. That's crazy slow. My machine was only a few minutes. The speed up makes sense suggestions make sense and lock files look correct.

`hedera` plugin had a test fail. Running again to see if it was a one off.

swcurran:
Integration tests failure doesn't bode well.  Failing in the Hedera plugin.

jamshale:
I can't see any library upgrade reason for `hedera` to be failing. Sometimes their test ledger has been broken. The easiest way to get this PR moving is to revert the lock file changes only for the `hedera` plugin. Then the integration tests for it will be skipped. 

It might come back or we can tag @AlexanderShenshin to see if he can have a look.

swcurran:
Claude-ing to see what's up. It's found an error so am testing it locally... will report soon.

swcurran:
OK -- I have a fix for that -- reproduced the fail locally, and the fix. I'm going to revert the Hedera plugin so that we can get this PR merged cleanly, and then will submit a Hedera PR with the lock file and code change.

swcurran:
Probably will fail again -- I didn't revert the Hedera integration poetry.lock file.  Confirming and then will update the PR and create the new one.  Fun stuff!

swcurran:
redis-events failing as well.  Investigating...

swcurran:
Redis Events update removed. Will take the same approach as with Hedera -- a separate PR, although I don't know what the issue is yet.

swcurran:
Sigh -- still failing.  Looks like redis_events again. Just noticed there are three redis_events lock files, so reverting that one as well.

swcurran:
w00t!!  Passed!  @jamshale -- can you please review/approve/merge (if appropriate) and then I will follow up with PRs for the not updated plugins.

esune (approved):

PR #3182, opened 2026-06-08
Fixes Hedera plugin after aiohttp upgrade

swcurran:
The helpers.py file had a URL construction bug where create_wallet and _auth_get both used f"{self.base_url}/{path}" — but since path already starts with /, this produced double-slash URLs like http://holder:3001//multitenancy/wallet.

Previously aiohttp silently normalized these away, but a version bump stopped doing that, causing all requests to return 404. The fix removes the extra / separator (matching the already-correct pattern used by _auth_post), and adds the missing leading / to two _auth_get call sites that were relying on the separator being there.

Found by Claude Code testing.


swcurran:
FYI -- @AlexanderShenshin @tajang97 -- nice to get your approval on this before merging. Pretty minor, I think.

AlexanderShenshin (approved):

WadeBarnes (approved):

PR #3184, opened 2026-06-09
Update redis_events/docker/services poetry lock and fix appled for uvicorn change

swcurran:
Upgrade of uvicorn was failing for poetry lock update in redis_events/docker/services. Fixed using Claude as follows:

fix(redis_events): replace uvicorn.run() with Server.serve() for uvicorn 0.49.0 compatibility

Uvicorn 0.49.0 changed its internal run() to use asyncio.run() instead of loop.run_until_complete(). This breaks the nest_asyncio pattern used in status_endpoints.py, which calls uvicorn.run() synchronously from within an async function.

Fix: replace uvicorn.run() with await uvicorn.Server(config).serve(), which is the correct async pattern. Also removes the now-unnecessary nest-asyncio dependency.

Includes the Dependabot uvicorn bump (0.35.0→0.49.0) in redis_events/docker/services/poetry.lock.

Closes #3165

- **build(deps): bump uvicorn in /redis_events/docker/services**
- **Update redis_events/docker/services poetry lock fix appled for uvicorn change**


swcurran:
Looks like this solves the failing integration tests from just updating the lock file for the redis_events/docker/services.  Weird...

jamshale (approved):

PR #3223, opened 2026-06-16
fix: update frontend Dockerfile

thiagoromanos:
Just fixing the oid4vc demo. The frontend after the libs updates now requires node 20 instead of 18

### DIFF
--- a/oid4vc/demo/frontend/Dockerfile
+++ b/oid4vc/demo/frontend/Dockerfile
@@ -1,4 +1,4 @@
-FROM node:20
+FROM node:22
 WORKDIR /app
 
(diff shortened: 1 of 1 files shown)


swcurran:
I'm tempted to just merge, but @timbl-ont or @weiiv -- you should confirm.

timbl-ont:
@thiagoromanos can you change to node 22. It is the LTS version. Thanks for submitting this PR.

thiagoromanos:
> @thiagoromanos can you change to node 22. It is the LTS version. Thanks for submitting this PR.

Done! Updated to node 22. Thanks for reminding me of that

swcurran (approved):

PR #3226, opened 2026-06-18
Follow up dependabot PR and updates to repo_manager.py to streamline the weekly process

swcurran:
Found a new pattern in the dependabot PRs that recent updates didn't address -- where a change to pyproject.toml files must be made in concert with the poetry.lock update. The change in this PR is an example of that.

Also used Claude to  update the repo_manager.py to improve the workflow:

- Changed option 7 to NOT take a list of libraries, but rather to generate the list itself, as was added recently with a second script.  New options:
  - `python repo_manager.py 7` - scan the dependabot PRs and list the ones to be updated.
  - `python repo_manager.py 7 --debug` - same as no parameter, but with more details.
  - `python repo_manager.py 7 --apply` - use the scanned list to update all the poetry.lock files for the libraries in question.
  - Any other -- print a message about the new behaviour, so people using the old mechanism realize it won't work.
- Added option 9 to handle the scenario of pyproject.toml files that have to be updated. Run with no parameters to get the list of updates and which pyproject.toml files will be updated, add `--apply` to make the pyproject.toml updates and regenerate the associated poetry.lock files.
  - Note that `--apply` is only done if the old version in the dependabot matches the version in the pypojrect.toml. If not, the "no parameter" option lists that, the "--apply" option ignores them.

The toml/lock files in this PR were generated using option 9.

I have not yet used the option 7 change yet, but will do so and adjust as needed with next week's updates.

FYI - @jamshale

I have not used the new optoon


WadeBarnes (approved):

WadeBarnes:
Will defer to @jamshale for final review.

jamshale (approved):
Cool! some complex logic but looks useful 👍

PR #3252, opened 2026-06-22
Weekly Update - 20260922

swcurran:
Weekly Update for June 22, 2026.

Updates dependencies:

- form-data
- google-auth
- pytest
- ruff
- undici
- pytest: "^9.0.3" -> "^9.1.1"
- ruff: "^0.15.12" -> "^0.15.18"
- ruff: "^0.15.14" -> "^0.15.18"

Also had an update to repo_manager.py because I crossed wires with Claude. I had first had Claude generate a Python script to scan dependabot PRs for upgrade, and then asked that it be integrated into repo_manager.py. What I didn't realize was that the update referenced the initial script, and I deleted that script after testing but before submitting the PR. Now fixed -- the code from the script is now within repo_manager.py.  This is why AI code for things that matter needs -- like things that go into production need a to be better understood by the submitter!!!  My fault!

Should be good to go for this week.  We'll see if any of the integration tests fail, and if any dependabot PRs remain open after this is merged.


swcurran:
Failure seems to because of (lack of) access to https://sandbox.bcvh.vonx.io.  It seems to be accessible now, so re-running the failed tests.

jamshale (approved):

PR #3268, opened 2026-06-29
Weekly Update - 20260629

swcurran:
Updates for this week:

- aioresponses
- google-api-python-client
- google-auth
- ruff: "^0.15.18" -> "^0.15.20"

Not updated was the file/lock file: `oid4vc/auth_server/pyproject.toml` as it is currently on a different version `"^0.13"`.

@timbl-ont, @weiiv can you check on that file ^^^^.


WadeBarnes (approved):

swcurran:
Looks like the integration test failure was because the server sandbox.bcvh.vonx.io  was unavailable.  @WadeBarnes -- can you take a peek at that?  @PatStLouis -- should we come up with an alternative for these tests. Same pattern the last few weeks.  Re-running and will likely pass now.

PR #3273, opened 2026-07-06
Weekly Dependabot Update -- 20260706

swcurran:
Weekly udpate.

- google-auth
- ruff
- typing-extensions
- uvicorn

Question: In this PR, we find that there are multiple pyproject.toml files that have dependencies on (in this case uvicorn) that have different supported versions string (e.g. `">=0.38,<0.51"` and `"^0.50.2"`). Would it be helpful/make sense to make those strings consistent across all of the files?  I can imagine that the settings might have been done on purpose, or that it was just done as a copy of another pyproject.toml file that had a different dependency, and it makes no difference.  Does consistency across files matter?

### DIFF
--- a/basicmessage_storage/poetry.lock
+++ b/basicmessage_storage/poetry.lock
@@ -2747,30 +2747,30 @@ test = ["hypothesis (>=6.22.0,<6.108.7)", "pytest (>=7.0.0)", "pytest-xdist (>=2
 
 [[package]]
 name = "ruff"
-version = "0.15.20"
+version = "0.15.21"
 description = "An extremely fast Python linter and code formatter, written in Rust."
 optional = false
 python-versions = ">=3.7"
 groups = ["dev"]
 files = [
-    {file = "ruff-0.15.20-py3-none-linux_armv6l.whl", hash = "sha256:00e188c53e499c3c1637f73c91dcf2fb56d576cab76ce1be50a27c4e80e37078"},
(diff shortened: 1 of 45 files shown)


swcurran:
Updated for another week of updates.  @jamshale @esune -- can I get an approval on this one?  Thanks!

jamshale:
It seems like the tests are consistently failing for `oid4vc` because a poetry install issues. I think it's possibly the `type-extensions` upgrade in the auth-server.

swcurran:
The failure (per AI) is because the Dockerfile is loading the source code for the Indicio isomdl-uniffi library, and its build process recently changed to include [zip](https://ziglang.org/].  To address, I went ahead with an update to the dockerfile to load zig into the image before the build.

@timbl-ont @weiiv -- please be aware of the change to the dockerfile as part of this PR. I can back it out if needed.  The other option was to pin the library to a specific version, which you might still want to do. I'm assuming that minor improvements and spec alignment are still happening, so elected to add zig vs. pin to an old version.  You might also want to encourage Indicio (@mepeltier perhaps?) to publish the library as an artifact so that we don't need to build it here.

Let's see if the tests pass now...

jamshale (approved):

esune (approved):

timbl-ont:
@swcurran looks good. Build is working and mdoc credentials can be issued.

The build phase for isomdl-uniffi is temporary, mostly to support macOS arm64. I believe that the repo is in the process of being moved to OWF. Once there, it would be good to get python wrappers published. FYI @mepeltier

PR #3290, opened 2026-07-15
OID4VC - use mdliso-mdoc lib from indicio

thiagoromanos:
The indicio lib for mso_mdoc has a whl package published, so there's no need to compile the lib inside the Dockerfile.

### DIFF
--- a/oid4vc/docker/Dockerfile
+++ b/oid4vc/docker/Dockerfile
@@ -24,32 +24,6 @@ RUN mkdir mso_mdoc/isomdl
 COPY pyproject.toml poetry.lock README.md ./
 RUN poetry install --without dev --all-extras
 
-#build the isomdl library from indicio. Tempory solution until we have a wheel on pypi with linux arm support
-RUN apt-get update && apt-get install -y git build-essential && apt-get clean
-RUN git clone https://github.com/Indicio-tech/isomdl-uniffi.git
-RUN curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y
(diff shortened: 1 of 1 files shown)


jamshale (approved):

timbl-ont:
@thiagoromanos The published whl does not support MacOS arm. That was the reason for the build in the dockerfile. In the future once the repo is moved to OWF we can work on publishing for all platforms. For now this workaround is required.

The PR is also hardcoded for Linux. Can we revert until all platforms are available?

timbl-ont:
Sorry I was not too clear on the above comment. 

It had been some time since the original changes. The issue is arm and not the OS. Of course the VM is running linux - brain hiccup on my part. The problem is that if the published whl is used then the whole container needs to be x86. This does work under Rosetta but at the time I recall it was not working that well. I don't recall the specific issues.

In any case what brought me here is that it is not building at all right now. I tried a clean install and the whl was not found. I will log an issue for that.

thiagoromanos:
> Sorry I was not too clear on the above comment.
> 
> It had been some time since the original changes. The issue is arm and not the OS. Of course the VM is running linux - brain hiccup on my part. The problem is that if the published whl is used then the whole container needs to be x86. This does work under Rosetta but at the time I recall it was not working that well. I don't recall the specific issues.
> 
> In any case what brought me here is that it is not building at all right now. I tried a clean install and the whl was not found. I will log an issue for that.

That's odd, right now I could build and run the container fine. The issue I'm having right now is with the auth server, which is another issue. ps: I'm running if from a ubuntu host.

PR #3342, opened 2026-07-27
Weekly Update -- 20260727

swcurran:


swcurran:
The weekly ACA-Py plugins PR is again ready for review.  The ruff update to 16 caused some linting errors in the redis cache plugin. Evidently most of the other plugins lock down the style/checks done by ruff, but this plugin doesn't.  Rather than locking it down as well (which requires more extensive code changes), Claude proposed just addressing the minor formatting issues flagged by ruff in the code.  Perhaps it might be worth someone updating the pyproject.toml file for that plugin -- and deal with the resulting required code updates.

For now -- this is likely ready for review and approval.  The integration tests fail on the first run was likely an intermittent error. Still running as I write this.

jamshale (approved):

PR #3387, opened 2026-08-04
Dependabot updates 20260804

swcurran:
More updates for this week:

aiohttp
cryptography


jamshale:
Looks like there was a breaking change for `aiohttp` library in the `cheqd` plugin. It may just need the unit tests updated. We either need to make the code changes or we can revert the upgrade just for cheqd library.

swcurran:
@jamshale -- just did some Claude-ing and applied a fix to get the Cheqd tests to pass. No code change in the plugin, just an addition to the test script to enable compatibility with the latest aiohttp library.

Waiting to confirm passing.

swcurran:
Note README.md and the added file describe the fix and when it can be removed. And that even if we don't remove it -- all good when it is no longer needed.

jamshale:
Re-running. Appears to be an intermittent issue.

jamshale (approved):

PR #3391, opened 2026-08-04
Plugins Release PR for ACA-Py 1.6.1

swcurran:
PR manually created to update the Plugins for ACA-Py Release 1.6.1.

Process was:

- Search and replace all instances of `acapy-agent = { version = "~1.6.0", optional = true }` to 1.6.1
  - Found 16 instances in 16 pyproject.toml files.
  - That matches the expectations from the Plugin Release PR for 1.6.0 #2919
- Run `python repo_manager.py 6` to update all poetry lock files -- updated the 16 lock files associated with the pyproject.toml files.
- Copy/paste 1.6.0 section of RELEASES.md and update all references to 1.6.1

@jamshale -- anything to be done after this runs?

### DIFF
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -2,6 +2,41 @@
 
 > **Note:** Release notes are also published as [GitHub Releases](https://github.com/openwallet-foundation/acapy-plugins/releases). This file may be phased out in favor of the CHANGELOG and release tags; feedback welcome.
 
+## ACA-Py Release 1.6.1
+
+| Plugin Name | Supported ACA-Py Release |
+| --- | --- |
+|basicmessage_storage | 1.6.1|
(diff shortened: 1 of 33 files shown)


jamshale:
This almost looks right, but I don't see the update for acapy-agent in the lock files. Only the hash has changed.

swcurran:
I think it is OK -- please take another look.  I think what has happened is that prior updates to the lock file changed the ACA-Py version to 1.6.1 (although not sure why since the pyproject.toml should have not callowed changes like that). If you expand the full lock file, it does say the ACA-Py version is 1.6.1 -- in the current code files and in the updated lock files.

jamshale (approved):

PR #3414, opened 2026-08-19
Weekly Dependabot Updates 2026.08.24

swcurran:
Update this week for:

- ruff
- uvicorn
- google-api-python-client
- pygments

Because the Integration Tests have been failing -- possibly because of a heavy load on the test did:webvh server we are using, I've added a concurrency on the tests in the repo.  Not sure it is the right thing, but we'll give it a try and see if we can get passed this failing tests issue.


jamshale:
It seems like there's a problem unrelated to this PR. It's hard to tell if there's an issue with the webvh sandbox server these tests run against or if there was possibly a regression with acapy. The acapy version is pinned to last release which passed when we upgraded plugins to it, though, so it seems more likely something is up with the webvh server.

swcurran:
I'm kind of inclined to think it is an issue with the local GHA docker services for some reason.  However, that there is one sandbox error muddies things a bit.

Here are the last successful runs: 

- Aug 11, 5:30 PM PDT 21m 37s
- Aug 11, 6:44 AM PDT  21m 30s

Note the time to run -- 21m on the passes vs. 30+ on the fails.   That to me suggests it is a GHA-based services that are the problem, not the sandbox.

Trying another run to see if things are better today.

esune:
@PatStLouis do you have any input? There were some issues related to an unreachable Tails server that have been resolved, but the issues with the WebVH server persist. Is this a verison mismatch between plugin and deployed WebVH server perhaps?

PatStLouis (approved):

jamshale (approved):

PR #3473, opened 2026-09-08
Add a script to enable local testing of an ACA-Py RC Release

swcurran:
Adds a script that can be used in a local repo clone to test an ACA-Py Release Candidate Release. This
increases the confidence in an ACA-Py Release Candidate to make easier the decision to move to a final
release of a planned version.

Script added and section added to README.md on why/how to use the script.

The script was generated by Claude, with a manual review, adjustments and a test run.


jamshale (dismissed):
I won't block this because it is a useful tool that maintainers could use, but I think the same thing could have been done with a python script which would be more readable and fits with the python focus of the project.

jamshale (approved):
Nice 👍 This is a lot cleaner.

PR #3474, opened 2026-09-09
mDL Revocation Support

mepeltier:
# Summary

Adds mDL/mdoc revocation support end-to-end — issuance with an embedded IETF status-list claim, status-list publishing/updates from the demo UI, and presentation-time status checking, and adds [zrok](https://zrok.io/) as an alternative tunnel provider for the OID4VC demo (alongside the existing ngrok setup)

## What's included

**mDL/mdoc revocation**
- `isomdl_mdoc_sign` (`mdoc/issuer.py`) now accepts an optional `status` claim, embedded into the MSO at issuance time.
- `mdoc/utils.py` adds status-list resolution/decoding: fetches the published IETF status-list JWT, decodes the bitstring, and returns the credential's current status (or a fail-closed error string if the check can't be completed).
- `mdoc/cred_verifier.py` / `mdoc/pres_verifier.py` wire status checking into presentation verification.
- Demo frontend creates an IETF-type status list for mdoc credentials at issuance and can revoke/re-publish through the existing "update status" UI.

**zrok demo support**
- New `docker-compose-zrok.yaml` + `env.zrok.example` — a zrok-based alternative to the ngrok demo stack, for environments where ngrok isn't viable (e.g. requires a paid plan for stable URLs).
- `zrok-watchdog.sh` polls the demo's public URL and restarts the share if it drops.
- `docker/entrypoint.sh`, `auth_server/docker/entrypoint.sh`, and `frontend/entrypoint.sh` now skip ngrok tunnel-introspection when the relevant URL env vars (`OID4VCI_ENDPOINT`, `TENANT_ISSUER_BASE_URL`, etc.) are already set directly — existing ngrok setups are unaffected.
- Renamed `AUTHSERVER_NGROK_URL` → `AUTHSERVER_PUBLIC_URL` internally (container-facing only); `docker-compose.yaml` maps the existing host-side `AUTHSERVER_NGROK_URL` var to it, so no `.env` changes are needed for existing ngrok users.

## Docker build — heads up for reviewers

This branch intentionally diverges from `main`'s current `oid4vc/docker/Dockerfile`, which installs a prebuilt `isomdl-uniffi` wheel (`v0.1.0-indicio.1`). That wheel does **not** have status-claim support, which the revocation feature above depends on. This PR instead builds `isomdl-uniffi` from source against the `Indicio-tech/isomdl-uniffi#feat/mso-status-claim` branch. This does result in slow build times and dependency on a mutable GitHub URL, rather than a stable release artifact. PRs are open in [`spruceid/isomdl`](https://github.com/spruceid/isomdl/pull/140) and [`Indicio-tech/isomdl_uniffi`](https://github.com/Indicio-tech/isomdl-uniffi/pull/33) to add status-claim support in the upstream dependencies.

Once `isomdl-uniffi` cuts an official release wheel with status-claim support, this Dockerfile should revert to the wheel-install pattern `main` currently uses (which I plan on taking care of before merging this PR, assuming those PRs get merged quickly enough). Left a `NOTE` comment in the Dockerfile itself as a pointer for whoever does that, if I don't have a chance to.

Separately, `.github/workflows/pr-linting-and-unit-tests.yaml` still installs the old `v0.1.0-indicio.1` wheel for unit tests (unchanged by this PR) — fine today since no unit test exercises the new `status` parameter directly, but worth knowing if that changes.

## Testing

- `ruff check .` and `ruff format --check .` pass.
- `poetry run pytest` — 45 passed.
- Manually exercised the zrok demo flow (issuance, presentation, status update/revocation) end-to-end.

## Requests for Feedback
General feedback for the PR as a whole is appreciated, but here are some specific items that I'd appreciate feedback on:
- Preexisting `ngrok` demo setup
  - Recent changes to ngrok require a paid account for multiple tunnels, which the demo uses. The ngrok setup _should_ be unchanged, but confirmation from someone with the ability to test that would be appreciated
- Confirmation of `zrok` demo instructions
  - I've done my best to make sure the instructions are accurate, but since these are my changes and my setup, it's possible I made some changes that didn't land in the documentation
- Confirmation of mDL Revocation
  - The core feature of this PR is adding mDL revocation support, so a thorough review there, and confirmation that it works on someone else's machine, would be great

### DIFF
--- a/oid4vc/auth_server/docker/entrypoint.sh
+++ b/oid4vc/auth_server/docker/entrypoint.sh
@@ -40,29 +40,35 @@ TUNNEL_ENDPOINT=${TUNNEL_ENDPOINT:-http://ngrok:4040}
 WAIT_INTERVAL=${WAIT_INTERVAL:-3}
 WAIT_ATTEMPTS=${WAIT_ATTEMPTS:-10}
 
-liveliness_check () {
-        set -o pipefail
-        for CURRENT_ATTEMPT in $(seq 1 "$WAIT_ATTEMPTS"); do
-                # Use jq to check if the 'issuer' tunnel is available
-                if curl -sf "${1}/api/tunnels" | jq -e 'any(.tunnels[]; .name == "authserver" and .public_url != null)' > /dev/null; then
-                        break
-                else
-                        if [[ $CURRENT_ATTEMPT -ge $WAIT_ATTEMPTS ]]; then
-                                echo "Failed while waiting for 'issuer' tunnel in ${1}/api/tunnels"
-                                exit 1
+# If TENANT_ISSUER_BASE_URL is already provided (for example when using zrok),
+# skip the ngrok introspection and use the provided URL.
+if [[ -z "${TENANT_ISSUER_BASE_URL:-}" ]]; then
+        liveliness_check () {
+                set -o pipefail
+                for CURRENT_ATTEMPT in $(seq 1 "$WAIT_ATTEMPTS"); do
+                        # Use jq to check if the 'authserver' tunnel is available
+                        if curl -sf "${1}/api/tunnels" | jq -e 'any(.tunnels[]; .name == "authserver" and .public_url != null)' > /dev/null; then
+                                break
+                        else
+                                if [[ $CURRENT_ATTEMPT -ge $WAIT_ATTEMPTS ]]; then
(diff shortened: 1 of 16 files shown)


timbl-ont:
@mepeltier as an FYI there will be a PR to add cbor support to the status list plugin. With both 18013-5 R2 IETF token status list will be covered.

